PR #8, opened 2021-02-28
added basic vegetation mining

Flapperkewiet:
Added support for vegetation mining

Not sure if i put everything in the right place


hubastard:
I think we will need a better foundation for to deal with local planets before I can merge this. Because right now this assumes that all the players are on the same localPlanet, which might not be the case. I had a discussion with @mushroom about that earlier the discussion is here if you want to follow along: https://github.com/hubastard/nebula/discussions/9

I think that I will stop what I was doing to look into this first, else it will be a blocker for a lot of stuff.

hubastard:
In fact, I will be able to merge it right now since you specified the planetId in the packet. It should be good for now.

PR #15, opened 2021-03-02
Improve the player client side interpolation

hubastard:
I've improved the `RemotePlayerMovement` client side interpolation algorithm to make sure that we still have smooth player movement even with high latency connection.

With this implementation, during my tests, I had still pretty decent movement smoothness even in the 200-300ms latency range.

I've also added some default connection latency and drop packet simulation for a "standard" remote connection to make sure that we have a realistic testing environment.

### DIFF
--- a/NebulaModel/Utils/TimeUtils.cs
+++ b/NebulaModel/Utils/TimeUtils.cs
@@ -1,0 +1,14 @@
+﻿using System;
+
+namespace NebulaModel.Utils
+{
+    public class TimeUtils
+    {
+        static readonly DateTime UNIX_EPOCH = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        public static long CurrentUnixTimestampMilliseconds()
+        {
+            return (long)(DateTime.UtcNow - UNIX_EPOCH).TotalMilliseconds;
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


NebulaModel/Utils/TimeUtils.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+
+namespace NebulaModel.Utils
+{
+    public class TimeUtils
+    {
+        static readonly DateTime UNIX_EPOCH = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        public static long CurrentUnixTimestampMilliseconds()
```
Mushroom, line 9: `DateTimeOffset.Now.ToUnixTimeSeconds()` now exists :)

Bit of an easier option than doing all this.

NebulaServer/Server.cs
```diff
@@ -24,6 +24,11 @@ public Server()
             server = new NetManager(this)
             {
                 AutoRecycle = true,
+                // Settings below only works in DEBUG builds
+                SimulateLatency = true,
```
Mushroom, line 28: Might be better to gate this regardless of whether it only applies to debug or not, as it could confuse us elsewhere.

Mushroom (changes requested):

NebulaServer/Server.cs
hubastard: Yeah sure

NebulaModel/Utils/TimeUtils.cs
hubastard: Doesn't exist since we are in 3.5 :(

PR #16, opened 2021-03-02
Added handshake process for clients

Mushroom:
This is an initial draft for the handshaking prototype, feedback is welcome/encouraged.

Several things have changed:
* There is now a handshaking sequence at the start
* The game pauses for all other players while the other player loads in
* There is now a basic status popup that shows for various events such as a player joining or leaving, or during connection

If you want to test this, you **need** to change the IP address on **line 18** of the `Server.cs` file to reflect the local IP of your machine on your network. Suggestions welcome on how to make this less finicky.


Flapperkewiet:
Looks very impressive.

How do you suggest testing this since Sandboxie doesn't work anymore? Is it possible to run two instances of steam/DSP on the same steam account? Or do you buy a second copy of the game :D

Mushroom:
Yes, I used a virtual machine and steam set to offline mode on it initially :) but also I anticipate it will be useful in the future to test things with multiple steam accounts, so I did also buy it on another account.

hubastard:
I'll probably investigate if I could detect that the game is running in Sandboxie and change de save path at runtime

hubastard:
@Mushroom had you seen this https://sandboxie-plus.com/sandboxie/writefilepath/ ?

It looks like you can whitelist some folders that allow Sandboxie to write to.

Mushroom:
I have not - if you can get that working that would be amazing

Flapperkewiet:
> Yes, I used a virtual machine and steam set to offline mode on it initially :)

That will work fine for now, but I can foresee it breaking when we need to identify players rejoining a server they already played on, since this wil most likely be based on some sort of steam-id, but that is trouble for later.

Flapperkewiet:
After some quick looking around it appears as if Sandboxie saves are in a seperate folder then the regular game:
Sandboxie: `C:\Sandbox\{Username}\DefaultBox\user\current\Documents\Dyson Sphere Program\Save`
Regular: `C:\Users\{Username}\Documents\Dyson Sphere Program\Save`

Some of the bepInEx folder is also seperated in Sandboxie:
Sandboxie: `C:\Sandbox\{Username}\DefaultBox\drive\C\Program Files (x86)\Steam\steamapps\common\Dyson Sphere Program\BepInEx`
Regular: `C:\Program Files (x86)\Steam\steamapps\common\Dyson Sphere Program\BepInEx`

So it is possible to view, for example, the logs of the sandboxie instance.

NebulaClient/MonoBehaviours/MultiplayerSession.cs
```diff
@@ -88,11 +99,24 @@ void Update()
...
 
         private void OnRemotePlayerJoined(RemotePlayerJoined packet)
         {
             PlayerManager.AddRemotePlayer(packet.PlayerId);
+            statusBox = UIMessageBox.Show("A new player is joining!", "A new player is joining, please wait while they load in", null, 0);
+            GameMain.Pause();
+
+            if(PlayerManager.WeAreMainPlayer)
+            {
+                GameSave.SaveCurrentGame("MPSYNCSTATE");
```
hubastard, line 117: I think we should use the `DSPGame.LoadFile` property to get the current save file name instead of creating our own. Otherwise, we won't be able to play 2 differents multiplayer games since when you will load the second game, it will override the first one.

For now since the `DSPGame.LoadFile` is empty when we start a new game, we could save it with a hardcoded  timesptamp.

NebulaClient/MonoBehaviours/MultiplayerSession.cs
```diff
@@ -88,11 +99,24 @@ void Update()
...
+                Client.SendPacket(new InitialState(GameConfig.gameSaveFolder + "/MPSYNCSTATE.dsv"));
```
hubastard, line 118: We should use `GameSave.saveExt` instead of hardcoding the save file extension. If they ever change the extension to something else it will still work.

NebulaClient/MonoBehaviours/MultiplayerSession.cs
```diff
@@ -114,5 +138,107 @@ private void OnVegeMined(VegeMined packet)
...
+        {
+            GameMain.Resume();
+            if(statusBox != null)
+            {
+                statusBox.FadeOut();
+                statusBox = null;
+            }
+        }
+
+        private class InitialStateSyncUtils
```
hubastard, line 190: I would prefer to have this class in its own file for ex: "NebulaClient/Utils/InitialStateSyncUtils.cs"

NebulaClient/MonoBehaviours/MultiplayerSession.cs
```diff
@@ -114,5 +138,107 @@ private void OnVegeMined(VegeMined packet)
...
+                        isRunning = false;
+
+                        if (request.isNetworkError || request.isHttpError)
+                        {
+                            Debug.LogError("Failed to download!");
+                        }
+                        else
+                        {
+                            byte[] receivedBytes = request.downloadHandler.data;
+                            string path = GameConfig.gameSaveFolder + "INITIALMPSTATE.dsv";
```
hubastard, line 213: Should move the name of the save file into a constant that you could use in OnInitState() above.

NebulaServer/GameLogic/PlayerManager.cs
```diff
@@ -80,5 +127,37 @@ public void PlayerDisconnected(NebulaConnection conn)
...
+
+        public void PlayerSentInitialState(Player player, InitialState packet)
+        {
+            if(!player.IsMasterClient)
+            {
+                // Someone is doing something nefarious here
+                return;
+            }
+
+            stateServer = new HttpStateServer($"http://{ Server.ServerIP }:17432", packet.URI);
```
hubastard, line 154: We could use the same port since the HTTP server is TCP and the game server UDP. That would have the benefit to only need to port forward 1 port instead of 2.

NebulaModel/NebulaModel.csproj
```diff
@@ -79,6 +83,6 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>copy /Y "$(TargetPath)" "%25PROGRAMFILES(x86)%25\Steam\steamapps\common\Dyson Sphere Program\BepInEx\plugins\Nebula\$(TargetFileName)"</PostBuildEvent>
+    <PostBuildEvent>copy /Y "$(TargetPath)" "D:\Games\Steam\steamapps\common\Dyson Sphere Program\BepInEx\plugins\Nebula\$(TargetFileName)"</PostBuildEvent>
```
hubastard, line 86: @hubastard Note to myself to look into having a better solution for this.

NebulaClient/GameLogic/PlayerManager.cs
```diff
@@ -8,6 +8,7 @@ public class PlayerManager
         Dictionary<ushort, Player> remotePlayers;
         Dictionary<ushort, RemotePlayerModel> remotePlayerModels;
 
+        public bool WeAreMainPlayer { get; set; } = false;
```
hubastard, line 11: Should rename to `IsMasterClient` and you can omit the `= false;` since the value is false by default.

NebulaServer/Server.cs
```diff
@@ -13,6 +14,9 @@ namespace NebulaServer
 {
     public class Server : INetEventListener
     {
+        //TODO: Some sort of config file
+        public static String ServerIP = "192.168.88.31";
```
hubastard, line 18: Set it to 127.0.0.1 for now, I'll look into adding some configuration file for our server to deal with this.

hubastard (changes requested):

hubastard (approved):

PR #17, opened 2021-03-02
Initial player color changing code

Flapperkewiet:
This is the inital version for changing the player color.

Currently the the player chooses a random color when joining a game.

Any feedback is welcome.


NebulaClient/GameLogic/PlayerManager.cs
```diff
@@ -1,79 +1,84 @@
...
 namespace NebulaClient.GameLogic
 {
-    public class PlayerManager
-    {
-        Dictionary<ushort, Player> remotePlayers;
-        Dictionary<ushort, RemotePlayerModel> remotePlayerModels;
+	public class PlayerManager
+	{
+		Dictionary<ushort, Player> remotePlayers;
+		Dictionary<ushort, RemotePlayerModel> remotePlayerModels;
```
hubastard, line 11: @hubastard Note to self, I'll need to add a `editorconfig` file with the code formatting rules of my project to prevent issues like different tabs identation that we see here.

NebulaClient/GameLogic/PlayerManager.cs
```diff
@@ -1,79 +1,84 @@
...
+			LocalPlayer = new Player(localPlayerId);
 
-        public void AddRemotePlayer(ushort playerId)
-        {
-            if (remotePlayers.ContainsKey(playerId))
-            {
-                Log.Error($"RemotePlayerManager :: Already contains the playerId {playerId}");
-                return;
-            }
+			// For now give the player a random color
```
hubastard, line 26: Please add // TODO: in from of this comment so we can keep track of this

NebulaModel/Packets/Players/PlayerColorChanged.cs
```diff
@@ -0,0 +1,17 @@
+﻿using NebulaModel.DataStructures;
+
+namespace NebulaModel.Packets.Players
+{
+	public class PlayerColorChanged
+	{
+		public ushort PlayerId { get; set; }
+		public Float4 Color { get; set; }
```
hubastard, line 8: We should probably use a Float3 for the color since we don't need to send the Alpha channel. It will always be 1.

NebulaServer/Player.cs
```diff
@@ -12,6 +12,7 @@ public class Player
         public Float3 RootPosition { get; set; }
         public Float3 RootRotation { get; set; }
         public Float3 BodyRotation { get; set; }
+        public Float4 PlayerColor { get; set; }
```
hubastard, line 15: Store it as Float3 also

NebulaServer/Server.cs
```diff
@@ -99,5 +100,13 @@ private void OnVegeMinedUpdate(VegeMined packet, NebulaConnection conn)
             Player player = playerManager.GetPlayer(conn);
             playerManager.SendPacketToOtherPlayers(packet, player, DeliveryMethod.ReliableUnordered);
         }
+
+        private void OnPlayerColorChanged(PlayerColorChanged packet, NebulaConnection conn)
+		{
+            Player player = playerManager.GetPlayer(conn);
+            player.PlayerColor = packet.Color;
+            Console.WriteLine($"Player {player.Id} changed to color {packet.Color}");
```
hubastard, line 108: @hubastard Note to self, should add support of the Log class in the server code also to have the same API everywhere.

NebulaServer/Server.cs
```diff
@@ -99,5 +100,13 @@ private void OnVegeMinedUpdate(VegeMined packet, NebulaConnection conn)
...
+            playerManager.SendPacketToOtherPlayers(packet, player, DeliveryMethod.ReliableOrdered);
```
hubastard, line 109: This can be `ReliableUnordered` since its not that important.

NebulaServer/GameLogic/PlayerManager.cs
```diff
@@ -56,11 +57,15 @@ public Player PlayerConnected(NebulaConnection conn)
             {
                 // Make sure that each player that is currently in the game receive that a new player join so they can create its RemotePlayerCharacter
                 player.SendPacket(new RemotePlayerJoined(newPlayer.Id));
+                // IDEA: Maybe send the entire player data in a packet? 
+                // This way we won't need to send seperate packets for each property of the player
+                player.SendPacket(new PlayerColorChanged(newPlayer.Id, player.PlayerColor));
```
hubastard, line 62: Both send packet of color update should be in the mode `ReliableUnordered`

hubastard (changes requested):

NebulaServer/GameLogic/PlayerManager.cs
Flapperkewiet: Wouldn't that make it possible for the PlayerColorChanged packet to arrive before the RemotePlayerJoined packet?

NebulaClient/GameLogic/PlayerManager.cs
Flapperkewiet: So, i change the tabs to spaces right?

NebulaServer/GameLogic/PlayerManager.cs
Mushroom: Perhaps it would be better to wait until my changes are merged in, as I fix this issue with a handshake packet.

NebulaClient/GameLogic/PlayerManager.cs
hubastard: Yeah, but if you merge master in your branch you should be able to now click the 'clean' button in a file to reformat it using the new .editorconfig file.

NebulaServer/GameLogic/PlayerManager.cs
hubastard: I've merged all the other PRs into master if you want to update work code with it.

NebulaClient/GameLogic/PlayerManager.cs
Flapperkewiet: Do you mean right-clicking solution ==> Clean Solution?

Flapperkewiet:
I believe everything should be fixed now? Otherwise, let me know.

PR #20, opened 2021-03-03
Hand-mined vegetation synced across planets

sp00ktober:
As long as mined with bare (mecha) hands it should be synced to all players even on different planets or in space.

I have noticed some desync but im pretty sure that it was caused by placing buildings as that removes vegetation too.
Anyways this is a wrok in progress but has good results so far, if you think something should be done differently please tell me :)


NebulaClient/MonoBehaviours/MultiplayerSession.cs
```diff
@@ -112,7 +112,72 @@ private void OnPlayerAnimationUpdate(PlayerAnimationUpdate packet)
...
+                }
+            }
+            else
+            {
+                planet = GameMain.galaxy?.PlanetById(packet.PlanetID);
+                if(planet == null)
+                {
+                    return;
+                }
+            }
```
hubastard, line 132: Can't you just always do `GameMain.galaxy?.PlanetById(packet.PlanetID)` even if it's your LocalPlanet ?

NebulaPatcher/Patches/Dynamic/PlayerAction_Mine_Patch.cs
```diff
@@ -0,0 +1,248 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Reflection.Emit;
+using UnityEngine;
+
+namespace NebulaPatcher.Patches.Dynamic
+{ 
+	[HarmonyPatch(typeof(PlayerAction_Mine), "GameTick")]
+	class PlayerAction_Mine_Transpiler
```
hubastard, line 13: I'll need to read up on Harmony Transpiler because I'm not sure that I understand all of what's going on here. But could you make a folder "Transpilers" in the folder "Patches" and put this file in there.

NebulaClient/MonoBehaviours/MultiplayerSession.cs
sp00ktober: I guess that should be possible.

NebulaPatcher/Patches/Dynamic/PlayerAction_Mine_Patch.cs
sp00ktober: The `patch_findMinfingId()`, `patch_findMiningId2()`, `patch_findRemoveVege()`, `patch_findMineVein()` functions search the `PlayerAction_Mine::GameTick()` function for specific OpCodes to find the right location where to inject the custom code.

As for now this is in `PlayerAction_Mine::GameTick()` line 57, line 99, line 119, line 130.
I insert a delegate function in there that gets called and consumes an `int` from the stack and puts one back there when finished (so the normal code execution can continue).

This way i can access `miningId` and call `MultiplayerSession.instance.Client.OnVegetationMined()` without the need of a Prefix patch and the computation done in the `GameTick()` function because my code gets called by the native function itself when needed.

hubastard (approved):

PR #24, opened 2021-03-04
Host refactor + packet processors

hubastard:
This is not done yet, but feel free to leave some comments.

So, I've been refactoring the project to have a NebulaClient for all the remote player logic and a NebulaHost for all the Host / Server logic. I still have a couple of things to fix to be at the same state that we were. The main thing that need to be fixed is the fact that the host is no longer connecting as a client to the server, the host is the server. That means that a lot of logic in the PlayerManager classes is needed some work.


hubastard:
@Mushroom @sp00ktober @Flapperkewiet
Please make sure to read the [wiki](https://github.com/hubastard/nebula/wiki/Project-Structure) since I've updated the documentation to reflect the new project structure.

PR #28, opened 2021-03-05
Player color & SetReady fixed by adding a callback for when the game finishes loading

Flapperkewiet:


Flapperkewiet:
There seems to be a problem with the syncing of colors, don't merge yet.

NebulaHost/PacketProcessors/Session/SyncCompleteProcessor.cs
```diff
@@ -18,7 +19,12 @@ public SyncCompleteProcessor()
 
         public void ProcessPacket(SyncComplete packet, NebulaConnection conn)
         {
+
             Player player = playerManager.GetSyncingPlayer(conn);
+            if (player == null)
+            {
+                Log.Warn("Received a SyncComplete packet, but no player is joining.");
+            }
```
hubastard, line 27: Should probably return in this if too. Else, it will throw an exception below.

NebulaPatcher/Patches/Dynamic/Update_Patch.cs
```diff
@@ -0,0 +1,15 @@
+﻿using HarmonyLib;
+using NebulaModel.Logger;
+using UnityEngine;
+
+namespace NebulaPatcher.Patches.Dynamic
+{
+    [HarmonyPatch(typeof(DSPGame), "Update")]
+    class Update_Patch
```
hubastard, line 8: Please rename the file and class to `DSPGame_Patch`

NebulaWorld/Utils/UnityUtils.cs
```diff
@@ -0,0 +1,31 @@
+﻿using NebulaModel.Logger;
+using UnityEngine;
+
+namespace NebulaWorld.Utils
+{
+    public static class UnityUtils
```
hubastard, line 6: You should remove this and use the very useful plugin [RuntimeUnityEditor](https://github.com/ManlyMarco/RuntimeUnityEditor/releases) instead. Once installed, you only need to press F12 to open it in game and you will be able to inspect any GameObject, Components and it even have a runtime console to run custom code at runtime.

You just need to download it and put it in the BepInEx/Plugins folder next to where the Nebula plugin is.
image

NebulaWorld/SimulatedWorld.cs
```diff
@@ -75,15 +79,14 @@ public static void UpdatePlayerColor(ushort playerId, Float3 color)
 
         public static void UpdatePlayerColor(Transform transform, Float3 color)
         {
-            // Apply new color to each part of the mecha
-            Renderer[] componentsInChildren = transform.gameObject.GetComponentsInChildren<Renderer>(includeInactive: true);
+            Log.Info($"Changing color of {transform.gameObject.name} to {color}");
 
+            // Apply new color to each part of the mecha
+            Renderer[] componentsInChildren = transform.gameObject.GetComponentsInChildren<Renderer>(includeInactive: false);
```
hubastard, line 85: Is the change to `includeInactive: false` really something that you wanted ?

NebulaWorld/SimulatedWorld.cs
```diff
@@ -144,5 +147,17 @@ public static void MineVegetable(VegeMined packet)
             }
         }
 
+        public static void OnGameLoadCompleted()
+        {
+            if (initialized == false)
+                return;
+
+            Log.Info("Game has finished loading");
+            //Assign our own coler
```
hubastard, line 156: typo: "color"

NebulaWorld/SimulatedWorld.cs
```diff
@@ -144,5 +147,17 @@ public static void MineVegetable(VegeMined packet)
...
+            UpdatePlayerColor(GameMain.data.mainPlayer.transform,
+                new Float3(UnityEngine.Random.Range(0.0f, 1.0f), UnityEngine.Random.Range(0.0f, 1.0f), UnityEngine.Random.Range(0.0f, 1.0f)));
```
hubastard, line 158: Shouldn't you use the `LocalPlayer.Data.Color` instead of generating a new random color ?

NebulaPatcher/Patches/Dynamic/GameLoader_Patch.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [HarmonyPatch(typeof(GameLoader), "FixedUpdate")]
+    class GameLoader_Patch
+    {
+        public static void Postfix(int ___frame)
+        {
+            Log.Info("FixedUpdate: " + ___frame);
+            if(___frame >= 10)
+            {
+                SimulatedWorld.OnGameLoadCompleted();
```
hubastard, line 17: Will this be called multiple times after the frame 10 ?

hubastard (changes requested):

NebulaPatcher/Patches/Dynamic/GameLoader_Patch.cs
Flapperkewiet: No I changed it to 11, since thats when the GameLoader self destructs

NebulaWorld/SimulatedWorld.cs
Flapperkewiet: Nowhere else is the player assigned a random color (not as far as i can tell)

NebulaWorld/SimulatedWorld.cs
Flapperkewiet: Not necessarily, it was just in case something was disabled, but it seems it never happens, so ill remove it.

NebulaWorld/Utils/UnityUtils.cs
Flapperkewiet: You are only telling me this NOW? I have been breaking my head trying to figure out whats going on inside the game and writing functions like this.

NebulaWorld/SimulatedWorld.cs
hubastard: It's done right now for remote client in the `NebulaHost.PlayerManager.PlayerConnected()`
And for the Host itself, it is in the `NebulaHost.MultiplayerHostSession.StartServer()`

NebulaWorld/Utils/UnityUtils.cs
hubastard: Sorry, I added this info to the Wiki 😄

Flapperkewiet:
There is still a problem with the color system. When a client joins it gets send the world data as well as the player data in one big packet which gets handled in `HandshakeResponseProcessor`.  The problem is, the game is still loading when the playerdata gets applied and the players are being overwritten. I don't know how to solve this. Anyone any ideas?

hubastard:
can you push your last changes to the PR I'll checkout it on my side to test

Flapperkewiet:
already in "fixes"

hubastard:
@Flapperkewiet I think you renamed the wrong file between `GameLoader_Patch` and `Update_Patch`, when I checkout your PR it doesn't find the file `GameLoader_Patch`

image

NebulaWorld/NebulaWorld.csproj
```diff
@@ -64,6 +64,7 @@
     <Compile Include="Properties\AssemblyInfo.cs" />
     <Compile Include="RemotePlayerModel.cs" />
     <Compile Include="SimulatedWorld.cs" />
+    <Compile Include="Utils\UnityUtils.cs" />
```
hubastard, line 67: Make sure to remove this reference also

hubastard:
@Flapperkewiet Alright, I fix it on my side, it seems to work now. There is just 1 null ref that occurs during on the host that I need to look at.

image

Flapperkewiet:
Hmm i didn't get any, which funciton? the UpdatePlayerColor?

NebulaWorld/SimulatedWorld.cs
```diff
@@ -67,26 +71,39 @@ public static void UpdateRemotePlayerAnimation(PlayerAnimationUpdate packet)
 
         public static void UpdatePlayerColor(ushort playerId, Float3 color)
         {
-            RemotePlayerModel player;
-            if (!remotePlayersModels.TryGetValue(playerId, out player))
+            Transform transform;
+            RemotePlayerModel remotePlayerModel;
+            if (LocalPlayer.IsMasterClient)
```
hubastard, line 76: This should be `playerId == LocalPlayer.PlayerId` instead of IsMasterClient, else you will apply the remote player color to the host.

hubastard:
> Hmm I didn't get any, which function? the UpdatePlayerColor?

Right now I'm having a Null ref in the `NebulaHost.PacketProcessors.Players.PlayerColorChangedProcessor.ProcessPacket ` I guess it can't find the player

Flapperkewiet:
> 
> 
> > Hmm I didn't get any, which function? the UpdatePlayerColor?
> 
When i check to see if the player is null it works. But when the client leaves and rejoins i get a `MovementPacketProcessor` error on the host.

hubastard:
I think I know why, I'll try something

Flapperkewiet:
Is it possible the connection ids and player ids are different for the same player?

hubastard:
@Flapperkewiet Yes connection id should not be used anymore. Since the host also needs an id and is never connected to himself.

Flapperkewiet:
Are you fixing this? Then i can go to bed :)

hubastard:
yes haha

hubastard (approved):

PR #32, opened 2021-03-07
Add initial sync of factory to player when they try to load it

Mushroom:
This pull request means that clients request the factory from the host upon attempting to load it, whether that be an initial connection, or entering the solar system from another one. This does *not* encompass unloading of factories from memory when a solar system is left.

This is done by overriding the `GameData.GetOrCreateFactory()` and `PlanetModelingManager.RequestLoadPlanetFactory()` methods to prevent it from being generated locally, and only adding it to the queue to be loaded in when the host sends the data.

Note - as this uses `PlanetFactory.Import()`, as opposed to creating from scratch, production statistics will need a closer look when we look at updating them from the host, as the indexes may not match. It appears that the indexes are assigned on a first-come-first-served basis, so if people visit planets in a different order that could cause a mismatch.

### DIFF
--- a/NebulaClient/NebulaClient.csproj
+++ b/NebulaClient/NebulaClient.csproj
@@ -74,6 +74,7 @@
   <ItemGroup>
     <Compile Include="MultiplayerClientSession.cs" />
     <Compile Include="PacketProcessors\GameStates\GameStateUpdateProcessor.cs" />
+    <Compile Include="PacketProcessors\Planet\FactoryDataProcessor.cs" />
     <Compile Include="PacketProcessors\Planet\VegetationMinedProcessor.cs" />
     <Compile Include="PacketProcessors\Players\PlayerAnimationUpdateProcessor.cs" />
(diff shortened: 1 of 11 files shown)


Flapperkewiet:
Can you add compression like discussed [here](https://github.com/hubastard/nebula/discussions/31)?
The library used can be found [here](https://github.com/MiloszKrajewski/lz4net). It says its oudated, but I couldn't get the k4os one to work with .Net Framework 3.5 and I doubt it would make much of a difference.

Then all you need to do is wrap the BinaryStream with an lz4CompressionStream (or lz4DecompressionStream) ([code snippet](https://github.com/Flapperkewiet/nebula/blob/b8645c2df70537a801ba86548de8199fcd02642c/NebulaWorld/CompressionBenchmark.cs#L156-L169))

hubastard (approved):

hubastard:
@Flapperkewiet I'll merge this PR as is for now. But, I agree that we should take a look at having a standard way of converting applying lz4 compression to our data before we send it, but this can be done in a subsequent PR. 😄

PR #37, opened 2021-03-08
Add LZ4 compression to the factory sync

Mushroom:
This stems from a comment made on my previous PR.

Merging this should wait until discussion is complete here: #31 

This is implemened with the 8K algorithm as mentioned in that discussion.

### DIFF
--- a/NebulaHost/NebulaHost.csproj
+++ b/NebulaHost/NebulaHost.csproj
@@ -37,6 +37,9 @@
     <Reference Include="Assembly-CSharp">
       <HintPath>..\Libs\Assembly-CSharp.dll</HintPath>
     </Reference>
+    <Reference Include="LZ4, Version=1.0.15.93, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\lz4net.1.0.15.93\lib\net2\LZ4.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
     <Reference Include="System.Xml.Linq" />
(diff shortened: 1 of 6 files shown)


sp00ktober:
did you make some comparison between lz4 and normal speed?

Mushroom:
If you read the linked discussion, a comparison was made for various settings

sp00ktober:
ah my bad i did not scroll down :joy:

NebulaPatcher/Patches/Dynamic/GameData_Patch.cs
```diff
@@ -35,7 +36,8 @@ static bool Prefix(GameData __instance, PlanetFactory __result, PlanetData plane
             // Import the factory from the given bytes, which will have been gotten or created on the host by the original function
             __instance.factories[__instance.factoryCount] = new PlanetFactory();
             using (MemoryStream ms = new MemoryStream(factoryBytes))
-            using(BinaryReader br =  new BinaryReader(ms))
+            using (LZ4Stream ls = new LZ4Stream(ms, CompressionMode.Decompress))
```
Therzok, line 39: This lacks the BufferedStream.

NebulaHost/NebulaHost.csproj
```diff
@@ -37,6 +37,9 @@
     <Reference Include="Assembly-CSharp">
       <HintPath>..\Libs\Assembly-CSharp.dll</HintPath>
     </Reference>
+    <Reference Include="LZ4, Version=1.0.15.93, Culture=neutral, processorArchitecture=MSIL">
```
Therzok, line 40: Please use PackageReference.

NebulaPatcher/Patches/Dynamic/GameData_Patch.cs
Mushroom: ACK, will fix next patchset

NebulaHost/NebulaHost.csproj
Mushroom: I just added the package normally via VS, no idea why it did it this way. Will try to move it over in the new patchset.

Therzok: Also, I think that the LZ4 lib should be a separate mod, simply because it might break in case multiple mods ship lz4.

Mushroom: Will sort.

I think for now it's okay with the LZ4 lib, I imagine this mod is going to be setting on a lot of toes in regards to other mods with that amount of method patching that we do - everything is contained in the "Nebula" folder already within the plugins folder. If it becomes an issue, then we can fix it.

Therzok (approved):

Mushroom:
I would guess this is now ready for review as @Flapperkewiet has done the tests which look good.

sp00ktober:
you need to add a check for `SimulatedWorld.Initialized` in `PlanetModelingManager_Patch.cs` and `GameData_Patch.cs` like here https://github.com/hubastard/nebula/pull/38/commits/866111c7001a0957c2d451da8b87a2760fb0b3a6
else you cannot load a singleplayer world straight from booting the game (it does work after you exit out of a multiplayer game back to main menu afaik)

Mushroom:
Looks like if that patch gets merged in first this should be okay

sp00ktober (approved):
Looks good to me.

hubastard (approved):

PR #38, opened 2021-03-08
Added a way to host any single player game from within the in-game Esc Menu

hubastard:
- Added a Host Game button inside the Esc Menu (in-game pause menu). That way we can now start a server from within any single player save files.
- Change the code to make sure that the host will send it's current `GameDesc` info to make sure that the clients loads with the right seed.
- Added some fixes to our patches methods to make sure we don't run them if we are not in a multiplayer session (`SimulatedWorld.Initialized == false`)


sp00ktober:
maybe we could add some small text in the upper right corner (where the game version and steam name is displayed) to tell the player that he is in a multiplayer game?

hubastard:
I'll do the save and add the multiplayer text tonight to be able to close this.

PR #45, opened 2021-03-11
Switch LiteNetLib to WebSocketsSharp

Mushroom:
This pull request swaps out LiteNetLib in exchange for WebSocketsSharp, relating to discussion #40 

The serializer from LiteNetLib has been retained, but we may want to use a different one in the future.

Data sync for large factories is now *considerably* faster when testing this in combination with #38.

This is a big change, so feedback and suggestions are welcome.


hubastard (approved):

sp00ktober (approved):
Needed to reference the websocket-sharp project in NebulaModel too to build, but beside that looks good to me.
also #48 was noticable faster in comming back to normal.

hubastard:
Fixed my own comment. Ready for merge.

PR #52, opened 2021-03-19
Fix remote players position in space

### LINKED ISSUE #33: Players clumb together once localPlanet gets null
movementSpace
rotation is still applied but the stick together until localPlanet of atleast one of them gets non-null again.

hubastard:
This PR fixes #33.

Since the local player becomes the center of the universe while in space and the universe literally revolves around him, we need to make sure to convert the remote player position received to be relative to the local player coordinate system.

Each local player will now send their LocalPlanetId and two positions in the Movement Packet, the `UPosition` and the `LocalPlayerPosition`.
 - When the player is in space or not on the same planet: we use the `UPosition` of the player which is the REAL absolute position of the player in the universe.
 - When the players are both on the same planet: we use the `LocalPlayerPosition` which is the position relative to the player's local planet. We can't rely on the `uPosition` here since the `uPosition` isn't as precise as the `LocalPlayerPosition` for planetary movement and it was causing the remote players model to clip a lot with the planet surface.

This PR also include a `Debugging_Patch` which is only applied in DEBUG build. It lets us unlock technologies instantly to quickly unlock the sailing upgrades.

In a separate PR and after we have all the @sp00ktober fixes and changes merges I will probably look at handling remote player visibility states. Since we should only render remote players when they are on the same planet then us or when we are both in space and close to each other.

### DIFF
--- a/NebulaHost/MultiplayerHostSession.cs
+++ b/NebulaHost/MultiplayerHostSession.cs
@@ -58,7 +58,7 @@ namespace NebulaHost
             LocalPlayer.IsMasterClient = true;
 
             // TODO: Load saved player info here
-            LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
         }
 
         private void OnConnectionRequest(ConnectionRequest request)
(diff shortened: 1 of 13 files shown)


sp00ktober:
Just tested it with a friend. we dont clumb together when we reach `localPlanet == null` but it seems the other player speeds up really fast and then disappears.
Once the other player went back to the planet and sailed around it i could see him in space sailing around there (guess thats when we should hide him)

hubastard:
@sp00ktober I had an issue when there was one player on the planet and the other one in space. Its now fixed, you can also toggle the player so they become big pink sphere to be easier to spot the other player position in space. Use the F10 key to toggle it.

hubastard:
https://www.youtube.com/watch?v=01cilJmICnE&ab_channel=HubertGendron

PR #55, opened 2021-03-19
Added latency simulation settings with our WebSocket implementation.

hubastard:
Since `WebSocketSharp` does not have latency simulation built-in, I added my own implementation to be able to have a more realistic latency when testing locally.

- Latency simulation can be enabled or disabled in `MultiplayerHostSession` and `MultiplayerClientSession`
- Latency simulation is only enabled in `DEBUG` builds
- Latency simulation will be enabled by default in `DEBUG` builds for both the Host and Client and they have a combined latency that ranges from 40ms to 100ms at random for each packet sent.

Note: I've also cleaned up the `PendingPacket` processing to be done inside the `PacketProcessor` class to share the same logic between the `Host` and `Client` code.

### DIFF
--- a/NebulaModel/Networking/DelayedPacket.cs
+++ b/NebulaModel/Networking/DelayedPacket.cs
@@ -2,7 +2,7 @@
 
 namespace NebulaModel.Networking
 {
-    public class DelayedPacket
+    public struct DelayedPacket
     {
         public PendingPacket Packet { get; }
         public DateTime DueTime { get; }
--- a/NebulaModel/Networking/DelayedPacket.cs
+++ b/NebulaModel/Networking/DelayedPacket.cs
@@ -1,0 +1,16 @@
+﻿using System;
+
+namespace NebulaModel.Networking
+{
+    public class DelayedPacket
+    {
+        public PendingPacket Packet { get; }
+        public DateTime DueTime { get; }
+
+        public DelayedPacket(PendingPacket packet, DateTime dueTime)
+        {
+            this.Packet = packet;
+            this.DueTime = dueTime;
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


NebulaModel/Networking/DelayedPacket.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace NebulaModel.Networking
+{
+    public class DelayedPacket
+    {
+        public PendingPacket Packet { get; }
+        public DateTime DueTime { get; }
+
+        public DelayedPacket(PendingPacket packet, DateTime dueTime)
```
Therzok, line 10: This could possibly allocate a lot during debugging, so maybe make it a struct?

NebulaModel/Networking/Serialization/NetPacketProcessor.cs
```diff
@@ -26,6 +35,64 @@ public NetPacketProcessor(int maxStringLength)
...
+                }
+            }
+        }
+
+        [Conditional("DEBUG")]
+        private void ProcessDelayedPackets()
+        {
+            lock (delayedPackets)
+            {
+                var now = DateTime.Now;
```
Therzok, line 84:
```suggestion
                var now = DateTime.UtcNow;
```

NebulaModel/Networking/Serialization/NetPacketProcessor.cs
```diff
@@ -26,6 +35,64 @@ public NetPacketProcessor(int maxStringLength)
...
 
+        public void EnqueuePacketForProcessing(byte[] rawData, object userData)
+        {
+#if DEBUG
+            if (SimulateLatency)
+            {
+                lock (delayedPackets)
+                {
+                    PendingPacket packet = new PendingPacket(rawData, userData);
+                    DateTime dueTime = DateTime.Now.AddMilliseconds(simulationRandom.Next(SimulatedMinLatency, SimulatedMaxLatency));
```
Therzok, line 46:
```suggestion
                    DateTime dueTime = DateTime.UtcNow.AddMilliseconds(simulationRandom.Next(SimulatedMinLatency, SimulatedMaxLatency));
```

Therzok:
Nice!

PR #62, opened 2021-03-21
Dedicated Server

Baldie-dev:
Possibility to run the server in the dedicated mode without the usage of GPU or any rendering.
The dedicated server can be started from the command line by:
`steam.exe -applaunch 1366540 -batchmode -nographics -server -load "asd"`
Where "asd" is the name of the savefile. 

Multiple methods are forced not to be executed due to the dependency on the GPU device (they need some GPU computing of the shaders). Otherwise, very few changes were made to the existing code, mainly just to prevent spawning the local player if "-server" parameter is present.

I have tested it with 2 players using the existing save file and it was working fine.

### DIFF
--- a/NebulaHost/MultiplayerHostSession.cs
+++ b/NebulaHost/MultiplayerHostSession.cs
@@ -14,7 +14,7 @@ namespace NebulaHost
     public class MultiplayerHostSession : MonoBehaviour, INetworkProvider
     {
         public static MultiplayerHostSession Instance { get; protected set; }
-        public static bool isDedicated = false;
+        public static bool IsDedicated = false;
 
         private WebSocketServer socketServer;
 
@@ -50,10 +50,8 @@ namespace NebulaHost
             LocalPlayer.IsMasterClient = true;
 
             // TODO: Load saved player info here
-            if (!isDedicated)
-            {
-                LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
-            }
+            LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            LocalPlayer.IsDedicated = IsDedicated;
         }
 
         private void StopServer()
--- a/NebulaHost/MultiplayerHostSession.cs
+++ b/NebulaHost/MultiplayerHostSession.cs
@@ -14,6 +14,7 @@ namespace NebulaHost
     public class MultiplayerHostSession : MonoBehaviour, INetworkProvider
     {
         public static MultiplayerHostSession Instance { get; protected set; }
+        public static bool isDedicated = false;
 
         private WebSocketServer socketServer;
 
@@ -49,7 +50,10 @@ namespace NebulaHost
             LocalPlayer.IsMasterClient = true;
 
             // TODO: Load saved player info here
-            LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            if (!isDedicated)
+            {
+                LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            }
         }
 
         private void StopServer()
--- a/NebulaHost/PlayerManager.cs
+++ b/NebulaHost/PlayerManager.cs
@@ -34,9 +34,9 @@ namespace NebulaHost
             return new PlayerData[] { LocalPlayer.Data }.Concat(GetConnectedPlayers().Select(p => p.Data));
         }
 
-        public IEnumerable<PlayerData> GetAllPlayerIdsExcludingHost()
+        public IEnumerable<PlayerData> GetAllPlayerDataExcludingHost()
         {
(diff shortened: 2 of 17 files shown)


Baldie-dev:
There is some issue with DysonSwarm_Patch, I should definitely not prevent Init() method to be executed. I am working on the workaround for that.

NebulaHost/PlayerManager.cs
```diff
@@ -34,6 +34,11 @@ public IEnumerable<PlayerData> GetAllPlayerIdsIncludingHost()
             return new PlayerData[] { LocalPlayer.Data }.Concat(GetConnectedPlayers().Select(p => p.Data));
         }
 
+        public IEnumerable<PlayerData> GetAllPlayerIdsExcludingHost()
+        {
+            return new PlayerData[] { }.Concat(GetConnectedPlayers().Select(p => p.Data));
```
hubastard, line 39: You can only do: `return GetConnectedPlayers().Select(p => p.Data);`

NebulaPatcher/MonoBehaviours/NebulaBootstrapper.cs
```diff
@@ -7,10 +7,17 @@ namespace NebulaPatcher.MonoBehaviours
...
         private void Awake()
-        {
+        {
+            string[] args = System.Environment.GetCommandLineArgs();
+            for (int i = 0; i < args.Length; i++)
+                if (args[i] == "-server")
+                {
+                    MultiplayerHostSession.isDedicated = true;
+                    break;
+                }
```
hubastard, line 20: I would prefer if you would wrap the for loop body between curly braces.

NebulaHost/MultiplayerHostSession.cs
```diff
@@ -14,6 +14,7 @@ namespace NebulaHost
     public class MultiplayerHostSession : MonoBehaviour, INetworkProvider
     {
         public static MultiplayerHostSession Instance { get; protected set; }
+        public static bool isDedicated = false;
```
hubastard, line 17: You should put this flag in the `LocalPlayer` class that we you will be able to do checks like `LocalPlayer.isDedicated` instead of `LocalPlayer.Data == null`

NebulaWorld/SimulatedWorld.cs
```diff
@@ -180,15 +180,18 @@ public static void OnGameLoadCompleted()
...
             Log.Info("Game has finished loading");
 
-            // Assign our own color
-            UpdatePlayerColor(LocalPlayer.PlayerId, LocalPlayer.Data.Color);
-
-            // TODO: Investigate where are the weird position coming from ?
-            // GameMain.mainPlayer.transform.position = data.Position.ToUnity();
-            // GameMain.mainPlayer.transform.eulerAngles = data.Rotation.ToUnity();
+           //This check is required when executed in dedicated mode
+            if (LocalPlayer.Data != null)
```
hubastard, line 184: Use `!LocalPlayer.isDedicated` instead

NebulaWorld/SimulatedWorld.cs
```diff
@@ -112,7 +112,7 @@ public static void UpdatePlayerColor(ushort playerId, Float3 color)
             }
 
             // We changed our own color, so we have to let others know
-            if (LocalPlayer.PlayerId == playerId)
+            if (LocalPlayer.Data != null && LocalPlayer.PlayerId == playerId)
```
hubastard, line 115: Use `!LocalPlayer.isDedicated` instead

NebulaWorld/SimulatedWorld.cs
```diff
@@ -85,7 +85,7 @@ public static void UpdatePlayerColor(ushort playerId, Float3 color)
         {
             Transform transform;
             RemotePlayerModel remotePlayerModel;
-            if (playerId == LocalPlayer.PlayerId)
+            if (LocalPlayer.Data != null && playerId == LocalPlayer.PlayerId)
```
hubastard, line 88: Use `!LocalPlayer.isDedicated` instead

NebulaHost/MultiplayerHostSession.cs
```diff
@@ -49,7 +50,10 @@ public void StartServer(int port)
             LocalPlayer.IsMasterClient = true;
 
             // TODO: Load saved player info here
-            LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            if (!isDedicated)
+            {
+                LocalPlayer.SetPlayerData(new PlayerData(PlayerManager.GetNextAvailablePlayerId(), GameMain.localPlanet?.id ?? -1, new Float3(1.0f, 0.6846404f, 0.243137181f)));
+            }
```
hubastard, line 56: I think that we should still do the `LocalPlayer.SetPlayerData` even if the dedicated host does not have an actual player. Else the host and the first player that joins will have the same player id and it could cause some issues in the future.

Baldie-dev:
Thanks! I will start addressing the comments. There is still one major blocker:

All positions and velocities of sails are being computed on the GPU via compute shaders which cannot be done with the"`-nographics`" flag, since there is no GPU device available that would handle it. They are doing simulation of Newtonian motion based on the gravity of the star, positions, and velocities of the sails. I can spend some time making optimized simulation on the CPU (ironically that was part of my college thesis :D), but it would be still overkill.

One workaround is to ignore the motions and just remember the initial position and velocity in orbit that is generated when the railgun fires. And this data can be sent to clients and they will do their own simulations. But this needs to be tested, once the Dyson sphere synchronization will be finished.

Baldie-dev:
I will close it for now since it does not make sense to pull something that is not fully working yet. I will resume it, once the majority of synchronization is implemented (including the Dyson sphere) so I can fully test that everything can be done without GPU.

Mushroom:
My suggestion would be to reopen this and mark it as a draft instead, so we can provide continuous feedback

hubastard:
@Baldie-dev  We have created a Discord for the mod if you want to join: https://discord.gg/vPvPv2vh

Issue #48, opened 2021-03-15
Weird planet moving when loading factory data from server


sp00ktober:
https://user-images.githubusercontent.com/79335616/111137696-6498ac00-857f-11eb-91e1-411bc74dd67f.mp4

Afaik this is due to loading the `PlanetFactory` from the server. It does not happen for Player 1 (the hosting one) or if you disable this feature for the others.

The error that pops up in the corner is fixed by [this patch](https://github.com/sp00ktober/nebula/commit/e9d2f3c8302751564e9cda6ecbe425ed916344e8)

- It seems like the weird planet movement starts when `ArrivePlanet()` or `RequestLoadPlanetFactory()` (this is where we patch and the request gets triggered) gets called.
- It seems like planet comes back to normal after `LoadingPlanetFactoryMain()` finished atleast one of its steps (not sure if it comes back sometime after the last step or sometime inbetween).
- `ModelingPlanetMain()` is only called once when you load the game, so its unlikely that the cause is in there.
- `uPosition` of player and planet do not change significantly between the weird movement of the planet.
- `GameMain.mainPlayer.transform.position` does change from `0 0 0` before the request to some coordinates after the request, but does not change when the planet comes back to normal. I dont think its related to this either.

Tested on master branch.

Mushroom:
This might be related to the gametick syncing we do - possibly stale data is loaded from P1, then when the gametick is updated after the factory is loaded in, the planet snaps back to where it should be.

sp00ktober:
it is comming back to normal faster with #45

sp00ktober:
fixed by [this commit](https://github.com/sp00ktober/nebula/commit/3d70c0b8c245c002fb9c256f838449f4ecb87219)

the way its done now "locks" the player at the edge between space and the max radius of the planet where `localPlanet` would be set until the factory is loaded. As this happens pretty fast (well atleast for the planet we have here) this is nearly not noticable.

It gets noticable tho when you arrive with full speed, so we may need to load the factory way before the game would normally do it to make this more smooth.

Should this smoothing be part of this issue or can we close here?

hubastard:
Nice job, as of the smoothing fix I suggest to do it later when we will be in a polish phase. For now we can focus on having a working feature first.

sp00ktober:
[...]

PR #70, opened 2021-03-22
Syncing Player Effects and Sounds, partial fix for weird planet movement

sp00ktober:
Okay i know this is a huge patch, i will make them smaller in the future..

**Effects synced:**
- handmining sparkles and sound effect
- thrusters visuals and sound when: drifting, flying, sailing
- footsteps visuals and sounds
- footsteps collision with vegetation sounds
- landing sound effect
- warping visual and sound

**Bugfixes**
- fixes the red error seen in the video in #48 
- fixes the weird planet movement described in #48 
- fixes a crash on subsequent planet arrivals

**Known issues**
- when standing on a building (like the starting pot) thrusters are activated due to not touching the ground
- warping is not really visible as high speed sailing tends to make remote players invisible just because of the immense speed.
- requesting, exporting, compressing, decompressing and importing of `FactoryData` freezes the server and the client for the processing time
- i use the refstuf in three places, need to fix that.

There are quite some places where the code can be improved, please review it and make suggestions :)


NebulaWorld/SimulatedWorld.cs
```diff
@@ -54,6 +55,18 @@ public static void SpawnRemotePlayerModel(PlayerData playerData)
...
+            {
+                if (GameMain.localPlanet != null)
+                {
+                    LocalPlayer.SendPacket(new localPlanetSyncPckt(GameMain.localPlanet.id, true));
+                }
+                else
+                {
+                    LocalPlayer.SendPacket(new localPlanetSyncPckt(0, true));
+                }
+            }
```
hubastard, line 69: Not sure I understand why this is in the `SpawnRemotePlayerModel` function

hubastard (changes requested):

NebulaWorld/SimulatedWorld.cs
```diff
@@ -78,6 +91,16 @@ public static void UpdateRemotePlayerAnimation(PlayerAnimationUpdate packet)
...
+        }
+
+        public static void UpdateRemotePlayerLocalPlanetId(localPlanetSyncPckt packet)
+        {
+            if(remotePlayersModels.TryGetValue(packet.playerId, out RemotePlayerModel player))
+            {
+                player.localPlanetId = packet.planetId;
+                player.Effects.UpdatePlanet(packet.planetId);
             }
         }
```
hubastard, line 105: This function can be removed and the body moved to `UpdateRemotePlayerPosition` above since the `PlayerMovement` packet already contains the `LocalPlanetId`

NebulaPatcher/Patches/Dynamic/PlanetModelingManager_Patch.cs
```diff
@@ -5,6 +5,13 @@
 using System.Collections.Generic;
 using System.Linq;
 
+using System.Security;
+using System.Security.Permissions;
+using UnityEngine;
+
+[module: UnverifiableCode]
+[assembly: SecurityPermission(SecurityAction.RequestMinimum, SkipVerification = true)]
+
```
hubastard, line 14: I don't think that you still need `System.Security` and `SecurityPermission` since you now use reflection ?

NebulaWorld/MonoBehaviours/Remote/RemotePlayerMovement.cs
```diff
@@ -112,6 +120,31 @@ private void MoveInterpolated(Snapshot previous, Snapshot current, float ratio)
...
             Vector3 previousPosition = GetRelativePosition(previous);
             Vector3 currentPosition = GetRelativePosition(current);
+            float deltaPosition = Vector3.Distance(previousPosition, currentPosition);
+            Vector3 velocity = (previousPosition - currentPosition) / (previous.Timestamp - current.Timestamp);
+
+            // needed as its null for some reason when we arrive here for the first time
+            if(rootWarp == null)
+            {
+                rootWarp = rootTransform.GetComponent<RemoteWarpEffect>();
+            }
```
hubastard, line 130: Try renaming the `Awake()` method to `Start()` method instead you should not have initialization race condition anymore.

`Awake()` in your case is called as soon as you add the component to the gameObject.
`Start()` is called on the frame when a script is enabled just before any of the Update methods are called the first time.

NebulaWorld/RemotePlayerModel.cs
```diff
@@ -32,6 +33,7 @@ public RemotePlayerModel(ushort playerId)
                 // Add remote player components
                 Movement = PlayerTransform.gameObject.AddComponent<RemotePlayerMovement>();
                 Animator = PlayerTransform.gameObject.AddComponent<RemotePlayerAnimation>();
+                Effects = PlayerTransform.gameObject.AddComponent<RemotePlayerEffects>();
```
hubastard, line 36: This is the reason for the race condition in `RemotePlayerMovement`, your `RemotePlayerEffects` was added after the `RemotePlayerMovement` script, so when the `RemotePlayerMovement.Awake()` was called, there was no `RemotePlayerEffects` on the player gameobject.

But the right fix is to change the `RemotePlayerMovement.Awake()` to `RemotePlayerMovement.Start()`

hubastard:
Will merge and fix the remaining comments in the next commit on master

PR #71, opened 2021-03-22
Add loading of planet data prior to computing planet

Mushroom:
This gets the planet data from the host when a client tries to load a star system/planet, before the planet is computed on the client.

This means that the mesh is correctly rendered on the client when it loads a factory, so factories are no longer floating in the air or half buried underground, and are aligned correctly with the terrain.

It also takes care of the case where a player might enter a star system that has not yet been entered by the host, so data is generated.

### DIFF
--- a/NebulaModel/Packets/Planet/PlanetDataResponse.cs
+++ b/NebulaModel/Packets/Planet/PlanetDataResponse.cs
@@ -1,0 +1,24 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace NebulaModel.Packets.Planet
+{
+    public class PlanetDataResponse
+    {
+        public int[] PlanetDataIDs { get; set; }
+        public byte[] PlanetDataBytes { get; set; }
+        public int[] PlanetDataBytesLengths { get; set; }
+
+        public PlanetDataResponse() { }
+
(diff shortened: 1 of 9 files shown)


NebulaModel/Packets/Planet/PlanetDataResponse.cs
```diff
@@ -0,0 +1,24 @@
...
+        public int[] PlanetDataBytesLengths { get; set; }
+
+        public PlanetDataResponse() { }
+
+        public PlanetDataResponse(Dictionary<int, byte[]> planetData) 
+        {
+            this.PlanetDataIDs = planetData.Keys.ToArray();
+
+            // Can't use a jagged array because LNL serializer says no, so flattening and separate offset array it is
+            // TODO: Possibly register a type with the serializer instead of doing this manually
```
hubastard, line 19: We could go with a custom type or I could look into the LiteNetLib serialization code and add support for jagged arrays if it's something that we will need a lot.

hubastard (approved):

NebulaModel/Packets/Planet/PlanetDataResponse.cs
Mushroom: Let's leave it as is for now, and then if it comes up again we can add a type or update the serializer, it's not critical

Issue #60, opened 2021-03-20
Joining a game always respawn the landing pod for remote clients.


hubastard:
When a client joins a game where the landing pod is already destroyed, a landing pod is still created for the client the is joining the game.

Baldie-dev:
I can take this one. So the goal is to generate a landing pod just once for the host player, right?

hubastard:
The landing pod should be there for everyone in a new game. But once it's destroyed, any other player that joins should not see the pod.

Baldie-dev:
I believe this can be closed since my fix #72  was already merged with master. :)

PR #72, opened 2021-03-23
Fix for the landing pod bug

Baldie-dev:
This is a bug fix for the [issue 60](https://github.com/hubastard/nebula/issues/60).
The landing pod is already being synchronized as part of VegeData, so this fix just prevents spawning another landing pod on the client-side and also prevents calling "FlattenTerrain" method that was removing the received landing pod from the server.

Tested:

1. Host game in MP
2. When a client connects, he sees the landing pod.
3. Client disconnects
4. Host will mine and remove landing pod.
5. When a client connects, he does not see the landing pod anymore.


Baldie-dev:
#60

sp00ktober (approved):
Looks good to me.

hubastard (approved):
Looks good

Therzok (approved):

PR #76, opened 2021-03-24
Fix a few MSBuild re-evaluation issues

Therzok:
The code now doesn't run redundantly to generate the environment
It is no longer required to downgrade the dlls.
All references happen in a shared place.
Output directory is created automatically.

### DIFF
--- a/NebulaClient/NebulaClient.csproj
+++ b/NebulaClient/NebulaClient.csproj
@@ -34,9 +34,6 @@
     <WarningLevel>4</WarningLevel>
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
-  <PropertyGroup>
-    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-  </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


NebulaClient/NebulaClient.csproj
```diff
@@ -92,10 +57,12 @@
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\NebulaModel\NebulaModel.csproj">
+      <SpecificVersion>true</SpecificVersion>
```
Therzok, line 60: This is an annoyance that I plan to get rid of, but it'll have to do for now.

hubastard (approved):

hubastard:
Nice job, it's a lot cleaner 👌

Therzok:
I'll gradually migrate this towards SDK projects, similarly to how I've done in my repo.

Main advantage is that you don't have to downgrade the assemblies anymore.

Mushroom (approved):

Therzok:
Btw, feel free to merge, I'd like to open separate small PRs than have this become pretty big.

PR #82, opened 2021-03-25
add preview of build, fix planet arrival freeze

sp00ktober:
This syncs the preview of placed buildings before they are constructed to all players on the same planet.

It also changes the way we load the `PlanetFactory`. If the local player already has received the data we load the local copy of it instead of requesting a new one from the server which speeds up the process by a good amount.
When a player places a building it will be broadcasted to all players and as long as they have a local copy of the `PlanetFactory` they will update it accordingly.
If they dont have a local copy they will request the whole factory on the next arrival of the planet just as it was until now.


NebulaPatcher/Patches/Dynamic/GameData_Patch.cs
```diff
@@ -86,4 +87,69 @@ public static bool Prefix(GameData __instance, PlanetData planet)
...
+                {
+                    __instance.OnActivePlanetFactoryLoaded(planet);
+                }
+                else
+                {
+                    planet.LoadFactory();
+                    planet.onFactoryLoaded += __instance.OnActivePlanetFactoryLoaded;
+                }
+            }
+            planet.onLoaded -= __instance.OnActivePlanetLoaded;
```
Therzok, line 112: This should be under the null check, otherwise it will NRE.

Therzok: Also, onFactoryLoaded vs onLoaded? Was this meant to be done in the event handler?

NebulaWorld/SimulatedWorld.cs
```diff
@@ -123,7 +123,15 @@ public static void UpdatePlayerColor(ushort playerId, Float3 color)
...
         {
-            EntityManager.PlaceEntity(packet);
+            if (packet.isPrebuild && packet.planetId == GameMain.localPlanet?.id)
+            {
+                EntityManager.PlaceEntityPrebuild(packet);
+            }
+            else
+            {
+                // if this player is currently not on the planet where the building is placed then dont spawn a prebuild
+                EntityManager.PlaceEntity(packet);
```
hubastard, line 133: Wouldn't this place 2 of the same entity if the player is not on the same local planet? One for the prebuild and one for the real building?

NebulaWorld/Factory/EntityManager.cs
```diff
@@ -94,13 +119,166 @@ public static int[] MinerGetUsefullVeins(ref int[] tmp_ids, ref int veinCount, V
...
+                        break;
+                    }
+                }
+            }
+
+            // the following should care for factory loading.
+            BuildEntity(pos, rot, proto, packet.protoId, pData);
+            // if the factory is still null it means we are a client and have not loaded the factory yet
+            // thats why we exit here. factory will be synced on arrival
+            Debug.Log("1");
```
hubastard, line 277: You can probably remove this `Debug.log()`

NebulaPatcher/Patches/Dynamic/PlanetFactory.cs
```diff
@@ -0,0 +1,58 @@
...
+                    {
+                        return true;
+                    }
+                }
+            }
+            sendPacket.OnEntityPlaced(prebuild.protoId, prebuild.pos, prebuild.rot, true);
+            return true;
+        }
+    }
+    class sendPacket
```
hubastard, line 50: Class should start with a capital `SendPacket`

NebulaPatcher/NebulaPatcher.csproj
```diff
@@ -40,7 +40,7 @@
     <Compile Include="Logger\BepInExLogger.cs" />
     <Compile Include="MonoBehaviours\NebulaBootstrapper.cs" />
     <Compile Include="NebulaPlugin.cs" />
-    <Compile Include="Patches\Dynamic\BuildFinally_Patch.cs" />
+    <Compile Include="Patches\Dynamic\PlanetFactory.cs" />
```
Baldie-dev, line 43: I think the naming convention is to add "_Patch" at the end of the name, right?

hubastard (approved):

Issue #73, opened 2021-03-23
Sync Dyson Sphere object


Baldie-dev:
Sync the Dyson sphere and all its components when a player joins the server.

Current status:

- [x] Initial Dyson Sphere synchronization
- [x] Real-time Synchronization of Dyson Swarm orbits planning (Finished, PR ready)
- [x] Real-time Synchronization of Dyson Sphere structure planning (Finished, PR ready)
- [x] Synchronize Realtime Sail Bullets (Finished, PR ready)

I am creating this just for tracking purposes since I just start to work on it.

PR #83, opened 2021-03-26
Realtime Synchronization of Designing Dyson Sphere

Baldie-dev:
This is another part of the #73.
Following events were synchronized in real-time:

- Create new Dyson Swarm orbit
- Remove Dyson Swarm orbit
- Create new Dyson Layer
- Remove Dyson Layer
- Add Node to the layer
- Remove Node
- Add Frame between two nodes
- Remove Frame
- Add Shell between multiple nodes
- Remove Shell
- Bullet correction (when bullet from railgun is fired it is created on the client and server at the same time, due to the animation event from the railgun building. The server just sent a correction packet that will update some bullet parameters that are subject to randomness).

Video on [youtube](https://www.youtube.com/watch?v=T8VLE92g5JM)

Notes:

1. I have to add a packet queue for the Adding frames. When a player is placing nodes that are automatically creating frames from the previous node, it creates two packets (add a node, add frame). But sometimes "add frame" arrives sooner than "add node" which caused an issue. So I will put "add frame" to the queue and wait for the "add node".

2. When Node is being removed I have to manually put logic to also removed all attached frames and shells to the node.


NebulaClient/PacketProcessors/Universe/DysonSphereAddNodeProcessor.cs
```diff
@@ -0,0 +1,37 @@
...
+            //Try to add frames that failed due to the missing nodes
+            DysonSphereLayer dsl = GameMain.data.dysonSpheres[packet.StarIndex]?.GetLayer(packet.LayerId);
+            DysonSphereAddFramePacket queuedPacked;
+            for (int i = DysonSphere_Manager.QueuedAddFramePackets.Count - 1; i >= 0; i--)
+            {
+                queuedPacked = DysonSphere_Manager.QueuedAddFramePackets[i];
+                Log.Info($"Procesing queue, try to add frame {queuedPacked.NodeAId } - {queuedPacked.NodeBId} ");
+                if (dsl?.nodePool[queuedPacked.NodeAId]?.id != 0 && dsl?.nodePool[queuedPacked.NodeBId]?.id != 0)
+                {
+                    Log.Info($"find! added!");
```
Baldie-dev, line 29: I forgot to remove print statements. I will fix it.

Issue #81, opened 2021-03-25
Game pause on the host is not synced with clients


Baldie-dev:
When the host pauses the game (by pressing ESC) the game tick will stop which causes desync on all clients, since their game tick will continue increasing. I think it is needed to sync the game pause or completely remove pausing the game when navigating to the menu.

Baldie-dev:
I will take care of this:
- [ ] Prevent game pausing when players navigate to the menu (in progress)

hubastard:
After merging the PR, I've notice that I don't really like this approach. Preventing the pause to occurs currently prevents the actual EscMenu from showing up, which prevents the players from exiting the game or from changing their game settings.

PR #84, opened 2021-03-26
Prevent game pause

Baldie-dev:
This is a fix for issue #81.

Basically, it prevents the game to be paused via pressing ESC button and via the pause button in the Dyson design UI screen.
The downside is, that it is needed to use reflection to manually pause the game.


hubastard (approved):

hubastard:
After some test, I realized that this does not work properly. I've added a TODO in the `GameMain_Patch` and will look at fixing it later.

NebulaClient/PacketProcessors/Session/PlayerJoiningProcessor.cs
```diff
@@ -12,7 +13,7 @@ public class PlayerJoiningProcessor : IPacketProcessor<PlayerJoining>
         public void ProcessPacket(PlayerJoining packet, NebulaConnection conn)
         {
             SimulatedWorld.SpawnRemotePlayerModel(packet.PlayerData);
-            GameMain.Pause();
+            typeof(GameMain).GetField("_paused", BindingFlags.Instance | BindingFlags.NonPublic).SetValue(GameMain.instance, true);
```
Therzok, line 16: Hmmm, is Harmony referenced? This could use AccessTools, which would set this in an optimized manner.

Therzok: https://harmony.pardeike.net/api/HarmonyLib.AccessTools.html#HarmonyLib_AccessTools_FieldRefAccess__2___0_System_Reflection_FieldInfo_

PR #86, opened 2021-03-27
fix for NRE at StationComponent.InternalTickRemote

### LINKED ISSUE #59: NullReferenceException after joining the game with planetary stations
This exception is thrown every FixedUpdate() call to the player who has joined the game. Seems to be caused by the stations.
Error Message

```NullReferenceException: Object reference not set to an instance of an object
  at StationComponent.InternalTickRemote (Int32 timeGene, Double dt, Single shipSailSpeed, Single shipWarpSpeed, Int32 shipCarries, .StationComponent[] gStationPool, .AstroPose[] astroPoses, VectorLF3 relativePos, Quaternion relativeRot, Boolean starmap, System.Int32[] consumeRegister) [0x00000] in <filename unknown>:0 
  at PlanetTransport.GameTick (Int64 time, Boolean isActive) [0x00000] in <filename unknown>:0 
  at PlanetFactory.GameTick (Int64 time) [0x00000] in <filename unknown>:0 
  at GameData.GameTick (Int64 time) [0x00000] in <filename unknown>:0 
  at GameMain.FixedUpdate () [0x00000] in <filename unknown>:0
```

Steps to reproduce:

1. You can use mine [savegame](https://drive.google.com/file/d/1CTA6JhYSiIVhXN6szeprketV5ieXdcDm/view?usp=sharing)
2. Launch the game normally as a single-player and from [Esc] game menu host the game.
3. Join the game with the second player and observe the error.

sp00ktober:
This fixes #59 by requesting the needed `FactoryData` of Planets that contain ILS/PLS stations when joining the game.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
@@ -1,0 +1,21 @@
+﻿using HarmonyLib;
+using NebulaWorld;
+
+namespace NebulaPatcher.Patches.Dynamic
+{
+    [HarmonyPatch(typeof(GalacticTransport), "AddStationComponent")]
+    class GalacticTransport_Patch
+    {
+        public static void Postfix(int planetId, StationComponent station)
+        {
+            foreach(int i in LocalPlayer.PlanetIdsWithLogistics)
+            {
+                if(i == planetId)
+                {
+                    return;
+                }
+            }
+            LocalPlayer.PlanetIdsWithLogistics.Add(planetId);
+        }
+    }
+}
--- a/NebulaClient/PacketProcessors/Planet/FactoryDataProcessor.cs
+++ b/NebulaClient/PacketProcessors/Planet/FactoryDataProcessor.cs
@@ -2,6 +2,7 @@
 using NebulaModel.Networking;
 using NebulaModel.Packets.Planet;
 using NebulaModel.Packets.Processors;
+using NebulaModel.Logger;
 using NebulaWorld;
 
 namespace NebulaClient.PacketProcessors.Planet
@@ -11,11 +12,35 @@ namespace NebulaClient.PacketProcessors.Planet
     {
         public void ProcessPacket(FactoryData packet, NebulaConnection conn)
         {
+            bool factoryJustForImport = false;
+
             LocalPlayer.PendingFactories.Add(packet.PlanetId, packet.BinaryData);
 
-            lock (PlanetModelingManager.fctPlanetReqList)
+            for(int i = 0; i < packet.PlanetIdsWithLogistics.Length && !LocalPlayer.IsMasterClient; i++)
             {
-                PlanetModelingManager.fctPlanetReqList.Enqueue(GameMain.galaxy.PlanetById(packet.PlanetId));
+                // if we have no loading queued and have not already loaded the factory data from server.
+                PlanetData pData = GameMain.galaxy.PlanetById(packet.PlanetIdsWithLogistics[i]);
+                if (!LocalPlayer.PendingFactories.ContainsKey(pData.id) && pData.factory == null && !pData.factoryLoading && !LocalPlayer.requestedAllLogistics)
+                {
+                    Log.Info($"Requested factory for planet (ID: {packet.PlanetIdsWithLogistics[i]}) from host because it contains logistic towers");
+                    LocalPlayer.SendPacket(new FactoryLoadRequest(packet.PlanetIdsWithLogistics[i]));
+                }
+                else if(packet.PlanetId == packet.PlanetIdsWithLogistics[i] && i != 0)
+                {
+                    // if its data for a factory we requested because of Logistics just import its data but do not really load it.
+                    // planet id at index 0 should always be the planet we spawn on so load it.
+                    GameMain.data.GetOrCreateFactory(pData);
+                    factoryJustForImport = true;
+                }
+            }
+            LocalPlayer.requestedAllLogistics = true;
+
+            if (!factoryJustForImport)
+            {
+                lock (PlanetModelingManager.fctPlanetReqList)
+                {
+                    PlanetModelingManager.fctPlanetReqList.Enqueue(GameMain.galaxy.PlanetById(packet.PlanetId));
+                }
             }
         }
     }
--- a/NebulaHost/PacketProcessors/Planet/FactoryLoadRequestProcessor.cs
+++ b/NebulaHost/PacketProcessors/Planet/FactoryLoadRequestProcessor.cs
@@ -3,6 +3,7 @@ using NebulaModel.Attributes;
 using NebulaModel.Networking;
 using NebulaModel.Packets.Planet;
 using NebulaModel.Packets.Processors;
+using NebulaWorld;
 using System.IO;
 using System.IO.Compression;
 
@@ -25,7 +26,7 @@ namespace NebulaHost.PacketProcessors.Planet
                     factory.Export(bw);
                 }
 
-                conn.SendPacket(new FactoryData(packet.PlanetID, ms.ToArray()));
+                conn.SendPacket(new FactoryData(packet.PlanetID, ms.ToArray(), LocalPlayer.PlanetIdsWithLogistics.ToArray()));
             }
         }
     }
--- a/NebulaModel/Packets/Planet/FactoryData.cs
+++ b/NebulaModel/Packets/Planet/FactoryData.cs
@@ -1,15 +1,20 @@
-﻿namespace NebulaModel.Packets.Planet
+﻿using System;
+
+namespace NebulaModel.Packets.Planet
 {
     public class FactoryData
     {
         public int PlanetId { get; set; }
         public byte[] BinaryData { get; set; }
+        public int[] PlanetIdsWithLogistics { get; set; }
 
         public FactoryData() { }
-        public FactoryData(int id, byte[] data)
+        public FactoryData(int id, byte[] data, int[] PlanetIdsWithLogistics)
         {
             this.PlanetId = id;
             this.BinaryData = data;
+            this.PlanetIdsWithLogistics = new int[PlanetIdsWithLogistics.Length];
+            Array.Copy(PlanetIdsWithLogistics, this.PlanetIdsWithLogistics, PlanetIdsWithLogistics.Length);
         }
     }
 }
--- a/NebulaPatcher/NebulaPatcher.csproj
+++ b/NebulaPatcher/NebulaPatcher.csproj
@@ -40,6 +40,7 @@
     <Compile Include="Logger\BepInExLogger.cs" />
     <Compile Include="MonoBehaviours\NebulaBootstrapper.cs" />
     <Compile Include="NebulaPlugin.cs" />
+    <Compile Include="Patches\Dynamic\GalacticTransport_Patch.cs" />
     <Compile Include="Patches\Dynamic\PlanetFactory_Patch.cs" />
     <Compile Include="Patches\Dynamic\Debugging.cs" />
     <Compile Include="Patches\Dynamic\ArriveLeavePlanet_Patch.cs" />
--- a/NebulaPatcher/Patches/Dynamic/PlanetModelingManager_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/PlanetModelingManager_Patch.cs
@@ -5,7 +5,6 @@ using NebulaWorld;
 using System.Collections.Generic;
 using System.Linq;
 using NebulaModel.Packets.Universe;
-using UnityEngine;
 
 namespace NebulaPatcher.Patches.Dynamic
 {
--- a/NebulaWorld/LocalPlayer.cs
+++ b/NebulaWorld/LocalPlayer.cs
@@ -13,6 +13,8 @@ namespace NebulaWorld
         public static PlayerData Data { get; private set; }
         public static Dictionary<int, byte[]> PendingFactories { get; set; } = new Dictionary<int, byte[]>();
         public static Dictionary<PrebuildData, int> prebuildReceivedList { get; set; } = new Dictionary<PrebuildData, int>();
+        public static List<int> PlanetIdsWithLogistics { get; set; } = new List<int>();
+        public static bool requestedAllLogistics = false;
 
         private static INetworkProvider networkProvider;
 


Mushroom (changes requested):
As discussed in discord, I disagree with this approach - we should not be loading in all the factories, we should just get authoritative updates on logistics vessel positions/travels from the host

sp00ktober:
This will be a bigger PR again i guess. This is just because the fix for #59 turns out to be a working sync implementation for ILS and its ships.

- [x] Fix issue with prebuildings (buildings are sometimes unclickable after placed, prebuildings do only vanish after every building is placed)
- [ ] Sync ILS UI
- [ ] Sync item transport done by ships
- [ ] Sync ship rendering

NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
```diff
@@ -0,0 +1,127 @@
...
+                    Debug.Log(i + ": null");
+                }
+            }
+        }
+
+        [HarmonyPostfix]
+        [HarmonyPatch("RemoveStationComponent")]
+        public static void RemoveStationComponent_Postfix(GalacticTransport __instance, int gid)
+        {
+            if (__instance.stationRecycleCursor > 0)
```
Therzok, line 101: Shouldn't this be a message sent to the host to try and remove the station component?

Especially, since this should be reference counted per player, so you don't remove station components that are still in use?

NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                    __instance.stationRecycleCursor = 0;
+                }
+            }
+            int debug = 0;
+            if (__instance.stationRecycleCursor > 0)
+            {
+                Debug.Log((__instance.stationPool[__instance.stationRecycle[__instance.stationRecycleCursor - 1]] == null) ? "ITS NULL" : "ITS NOT NULL");
+                Debug.Log("recycle " + __instance.stationRecycleCursor);
+                debug = __instance.stationRecycle[__instance.stationRecycleCursor-1];
```
Therzok, line 46: Uhm, you shouldn't do -1 here. `stationRecycle[0]` is _never_ used. Only indices [1..capacity]

NebulaWorld/Logistics/StationUIManager.cs
```diff
@@ -0,0 +1,433 @@
...
+using NebulaModel.DataStructures;
+using NebulaModel.Networking;
+using UnityEngine;
+using UnityEngine.EventSystems;
+
+namespace NebulaWorld.Logistics
+{
+    public static class StationUIManager
+    {
+        private static ThreadSafeDictionary<int, List<NebulaConnection>> StationUISubscribers;
```
Therzok, line 14: This is gone in latest master.

NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
sp00ktober: decrement
it does access index 0 too :)

NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
sp00ktober: yes i probably should make this a message to the host too, thanks.

NebulaClient/PacketProcessors/Logistics/ILSAddStationComponentResponseProcessor.cs
```diff
@@ -0,0 +1,31 @@
...
+    [RegisterPacketProcessor]
+    public class ILSAddStationComponentResponseProcessor: IPacketProcessor<ILSAddStationComponentResponse>
+    {
+        public void ProcessPacket(ILSAddStationComponentResponse packet, NebulaConnection conn)
+        {
+            for(int i = 0; i < ILSShipManager.AddStationComponentQueue[packet.planetId].Count; i++)
+            {
+                if(ILSShipManager.AddStationComponentQueue[packet.planetId][i].shipDockPos == DataStructureExtensions.ToVector3(packet.shipDockPos))
+                {
+                    LocalPlayer.PatchLocks["GalacticTransport"] = true;
```
Therzok, line 21: There is a toggleswitch abstraction now that can be used for this.

NebulaModel/Packets/Logistics/ILSgStationPoolSync.cs
```diff
@@ -0,0 +1,129 @@
...
+                                    int[] shipIndex,
+                                    Double3[] shipPos,
+                                    Float4[] shipRot,
+                                    Float3[] shipVel,
+                                    float[] shipSpeed,
+                                    Float3[] shipAngularVel,
+                                    Double3[] shipPPosTemp,
+                                    Float4[] shipPRotTemp)
+        {
+            this.stationGId = new int[stationGId.Length];
```
Therzok, line 70: These could be `Array.Clone()`. Something like `this.stationGId = (int[])stationGId.Clone()`.

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        public void ProcessPacket(StationUIInitialSyncRequest packet, NebulaConnection conn)
+        {
+            StationComponent[] gStationPool = GameMain.data.galacticTransport.stationPool;
+            if(packet.stationGId < gStationPool.Length)
+            {
+                StationComponent stationComponent = gStationPool[packet.stationGId];
+                if(stationComponent != null)
+                {
+                    List<int> itemId = new List<int>();
```
Therzok, line 21: Is this sending data for entries that aren't set?

Therzok: Also, you can pre-allocate arrays here.

```csharp
var storage = storageComponent.storage;
var itemId = new int[storage.Length];
...

for (int i = 0; i < storage.Length; ++i) {
    var entry = storage[i];
    itemId[i] = entry.itemId;
    itemCountMax = entry.max;
    ..
}

File: NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs [hunk 7]
```diff
@@ -0,0 +1,72 @@
...
+                    List<int> itemCountMax = new List<int>();
+                    List<int> localLogic = new List<int>();
+                    List<int> remoteLogic = new List<int>();
+                    for(int i = 0; i < stationComponent.storage.Length; i++)
+                    {
+                        itemId.Add(stationComponent.storage[i].itemId);
+                        itemCountMax.Add(stationComponent.storage[i].max);
+                        switch (stationComponent.storage[i].localLogic)
```
Therzok, line 29: You can simply use `(int)localLogic`

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+                            case (ELogisticStorage.Supply):
+                                localLogic.Add(1);
+                                break;
+                            case (ELogisticStorage.Demand):
+                                localLogic.Add(2);
+                                break;
+                        }
+                        switch (stationComponent.storage[i].remoteLogic)
+                        {
+                            case (ELogisticStorage.None):
```
Therzok, line 43: Same here, `(int)remoteLogic`

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+                    }
+                    StationUIInitialSync packet2 = new StationUIInitialSync(packet.stationGId,
+                                                                            stationComponent.tripRangeDrones,
+                                                                            stationComponent.tripRangeShips,
+                                                                            stationComponent.deliveryDrones,
+                                                                            stationComponent.deliveryShips,
+                                                                            stationComponent.warpEnableDist,
+                                                                            stationComponent.warperNecessary,
+                                                                            stationComponent.includeOrbitCollector,
+                                                                            itemId.ToArray(),
```
Therzok, line 62: If you pre-allocate the arrays above, you don't need to ToArray() here.

NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
```diff
@@ -0,0 +1,79 @@
...
+                    stationComponent.warperNecessary = packet.warperNecessary;
+                    stationComponent.includeOrbitCollector = packet.includeOrbitCollector;
+                    for(int i = 0; i < packet.itemId.Length; i++)
+                    {
+                        stationComponent.storage[i].itemId = packet.itemId[i];
+                        stationComponent.storage[i].max = packet.itemCountMax[i];
+                        switch (packet.localLogic[i])
+                        {
+                            case 0:
+                                stationComponent.storage[i].localLogic = ELogisticStorage.None;
```
Therzok, line 42: Get rid of switch :p `localLogic = (ELogisticStorage)intValue;`

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
sp00ktober: Ah thanks, i should put the sending code inside the check.

NebulaWorld/Logistics/ILSShipManager.cs
```diff
@@ -209,13 +218,64 @@ public static void AddTakeItem(ILSShipItems packet)
...
+                Debug.Log(GameMain.galaxy.PlanetById(stationComponent.planetId).displayName + ": dir: " + stationComponent.workShipDatas[packet.origShipIndex].direction + " carry " + stationComponent.workShipDatas[packet.origShipIndex].itemId + "(" + stationComponent.workShipDatas[packet.origShipIndex].itemCount + ")");
+            }
+        }
+
+        private static void RefreshValuesUI(StationComponent stationComponent, int storageIndex)
+        {
+            UIStationWindow stationWindow = UIRoot.instance.uiGame.stationWindow;
+            if (stationWindow != null && (int)AccessTools.Field(typeof(UIStationWindow), "_stationId")?.GetValue(stationWindow) == stationComponent.id)
+            {
+                UIStationStorage[] stationStorageUI = (UIStationStorage[])AccessTools.Field(typeof(UIStationWindow), "storageUIs").GetValue(stationWindow);
```
Therzok, line 275: please cache these 3 AccessTools into FieldRef delegates. From the looks of it, these are called often, and they create allocations - possibly introducing more points for host/client to deync.

sp00ktober:
I leave it as it is now, the rest that is missing or thats still bugged will be covered by individual issues.
With this you should be able to play the game and make use of ILS/PLS and Orbital Collectors.
Both clients and host should be able to place the stations anywhere they like and configure them, add and remove items and objects (drones, ships, warpers).

NebulaPatcher/Patches/Dynamic/GalacticTransport_Patch.cs
sp00ktober: This is removed in current code, i added an issue for the error popping up when a client removes an ILS (#241 )

NebulaWorld/Logistics/StationUIManager.cs
sp00ktober: Im using the normal `Dictionary` there for now

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
sp00ktober: is it bad to use `ToArray()` in this situation?

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
sp00ktober: changed in next commit

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
sp00ktober: changed in next commit

NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
sp00ktober: changed in next commit :D

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
sp00ktober: preallocation is done in next commit :)

NebulaModel/Packets/Logistics/ILSgStationPoolSync.cs
sp00ktober: implemented like that in next commit.

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
Therzok: ToArray() creates a copy of the collection.

```csharp
var list = new List<int>(); // this uses an int[8] internally
for (int i = 0; i < source.Length; ++i) {
   // If the list array is full, it doubles it, so you get another allocation for
   // int[16], int[32], int[64], etc.
   // you also need to copy hte values to the new array
   list.Add(source[i]); 
}
var result = list.ToArray(); // another `array[list.Count]` allocation + copy
```

if you just do
```csharp
// just one allocation and copy
var result = (int[])source.Clone();

// What clone does
var result = new int[source.Length];
Array.Copy(source, result, source.Length);

// Similarly you have
var result = Array.ConvertAll(source, x => x);
```

sp00ktober: Thanks for the detail example, that makes it really clear 😄

NebulaWorld/Logistics/ILSShipManager.cs
sp00ktober: included in the next commit.

NebulaModel/Packets/Logistics/ILSShipData.cs
```diff
@@ -0,0 +1,43 @@
...
+        public int planetAStationGID { get; set; }
+        public int planetBStationGID { get; set; }
+        public int origShipIndex { get; set; }
+        public int warperCnt { get; set; }
+        public int stationWarperCnt { get; set; }
+
+        public ILSShipData() { }
+        public ILSShipData(bool IdleToWork, int planetA, int planetB, int itemId, int itemCount, int AGID, int BGID, int origShipIndex, int warperCnt, int stationWarperCnt)
+        {
+            this.IdleToWork = IdleToWork;
```
hubastard, line 19: `IdleToWork` should start with a lowerCase

NebulaModel/Packets/Logistics/ILSShipItems.cs
```diff
@@ -0,0 +1,21 @@
...
+        public bool AddItem { get; set; }
+        public int itemId { get; set; }
+        public int itemCount { get; set; }
+        public int origShipIndex { get; set; }
+        public int stationGID { get; set; }
+
+        public ILSShipItems() { }
+        public ILSShipItems(bool AddItem, int itemId, int itemCount, int origShipIndex, int stationGID)
+        {
+            this.AddItem = AddItem;
```
hubastard, line 14: `AddItem` should start with a lowercase

NebulaModel/Packets/Logistics/ILSSystemStorageSync.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace NebulaModel.Packets.Logistics
+{
+    public class ILSSystemStorageSync
+    {
+        public int[] stationGId { get; set; }
+        public int[] storageStationGId { get; set; }
```
hubastard, line 6: No constructor with array copy here ?

NebulaPatcher/NebulaPlugin.cs
```diff
@@ -43,6 +44,14 @@ private void InitPatches()
...
             {
                 var assemblies = AppDomain.CurrentDomain.GetAssemblies();
+                //Environment.SetEnvironmentVariable("MONOMOD_DMD_TYPE", "mb");
+                //Environment.SetEnvironmentVariable("MONOMOD_DMD_DUMP", "./mmdump");
+                /*
+                foreach (FileInfo file in new DirectoryInfo("./mmdump").GetFiles())
+                {
+                    file.Delete();
+                }
+                */
```
hubastard, line 54: Maybe we could add another build target that enables this

NebulaWorld/Logistics/StationUIManager.cs
```diff
@@ -0,0 +1,464 @@
...
+                    UpdateStorageUI(packet);
+                }
+                else
+                {
+                    UpdateSettingsUI(packet);
+                }
+            }
+        }
+
+        private static void UpdateSettingsUIBackground(StationUI packet, PlanetData pData, int stationGId)
```
hubastard, line 105: This method seems heavy and it looks like there is some code copied from the game. We should probably look if we can simplify it or call the actual game code instead.

NebulaModel/Packets/Logistics/ILSSystemStorageSync.cs
sp00ktober: Ah yes, i started to work on something but made an individual issue out of it. Im gonna remove this file for now.

NebulaPatcher/NebulaPlugin.cs
sp00ktober: Im not sure how this would be done :/

NebulaClient/PacketProcessors/Logistics/ILSAddStationComponentResponseProcessor.cs
sp00ktober: I *think* this is not the usecase here. What i need to do is set a boolean that i can check for in my Prefix patches to determine if they should run the patch code or the vanilla code in order to avoid endless loops.

NebulaWorld/Logistics/StationUIManager.cs
sp00ktober: I am calling the vanilla game functions in `UpdateSettingsUI()`, but since they belong to `UIStationWindow` i cant call them in `UpdateSettingsUIBackground()` since the window is closed at that time.

I could use reverse patching to get a modified version of the original method which i then could use there, but in the end it would be more complicated code including transpilers to get the same done.

I can see how i can make the method a bit easier to read.
basically its the functionality of 
grafik

Therzok (approved):

NebulaHost/PacketProcessors/Logistics/ILSgStationPoolSyncRequestProcessor.cs
```diff
@@ -0,0 +1,128 @@
...
+        public void ProcessPacket(ILSRequestgStationPoolSync packet, NebulaConnection conn)
+        {
+            Player player = playerManager.GetPlayer(conn);
+            if(player == null)
+            {
+                player = playerManager.GetSyncingPlayer(conn);
+            }
+            if (player != null)
+            {
+                List<int> stationGId = new List<int>();
```
Therzok, line 27: This is still ulsing lists and could be using arrays though.

sp00ktober: ups

NebulaPatcher/NebulaPlugin.cs
hubastard: I'll look at doing this later

hubastard (approved):

NebulaHost/PacketProcessors/Logistics/ILSgStationPoolSyncRequestProcessor.cs
```diff
@@ -0,0 +1,149 @@
...
+                int[] stationId = new int[countILS];
+                Float3[] DockPos = new Float3[countILS];
+                Float4[] DockRot = new Float4[countILS];
+                int[] planetId = new int[countILS];
+                int[] workShipCount = new int[countILS];
+                int[] idleShipCount = new int[countILS];
+                ulong[] workShipIndices = new ulong[countILS];
+                ulong[] idleShipIndices = new ulong[countILS];
+
+                int[] shipStationGId = new int[countILS * 10];
```
Mushroom, line 52: I feel like a const as opposed to a fixed "10" would work better here, in the event someone wants compatibility with a mod that increases capacities, or the game has ILSes tweaked in the future - I see it repeated further down also

NebulaHost/PacketProcessors/Logistics/StationSubscribeUIUpdatesProcessor.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    [RegisterPacketProcessor]
+    class StationSubscribeUIUpdatesProcessor: IPacketProcessor<StationSubscribeUIUpdates>
+    {
+        public void ProcessPacket(StationSubscribeUIUpdates packet, NebulaConnection conn)
+        {
+            if (packet.subscribe)
+            {
+                StationUIManager.AddSubscriber(packet.stationGId, conn);
+                // TODO: sync current state to client
```
Mushroom, line 17: Intentional TODO here?

NebulaModel/Networking/Serialization/NetPacketProcessor.cs
```diff
@@ -22,8 +22,8 @@ private static class HashCache<T>
         private Queue<PendingPacket> pendingPackets = new Queue<PendingPacket>();
 
         public bool SimulateLatency = false;
-        public int SimulatedMinLatency = 20;
-        public int SimulatedMaxLatency = 50;
+        public int SimulatedMinLatency = 15;
```
Mushroom, line 25: Should these values be changed? They are unrelated to ILSes

NebulaPatcher/NebulaPlugin.cs
```diff
@@ -43,6 +44,14 @@ private void InitPatches()
             try
             {
                 var assemblies = AppDomain.CurrentDomain.GetAssemblies();
+                //Environment.SetEnvironmentVariable("MONOMOD_DMD_TYPE", "mb");
```
Mushroom, line 47: Do not leave commented out code in the codebase

NebulaPatcher/Patches/Transpilers/PlayerAction_Mine_Patch.cs
```diff
@@ -30,7 +30,7 @@ static IEnumerable<CodeInstruction> GameTick_Transpiler(IEnumerable<CodeInstruct
                     new CodeMatch(OpCodes.Ldarg_0),
                     new CodeMatch(OpCodes.Ldfld),
                     new CodeMatch(i => i.opcode == OpCodes.Callvirt && ((MethodInfo)i.operand).Name == "GetVegeData"))
-                .Insert(Transpilers.EmitDelegate<Func<int, int>>(miningId =>
+                .Insert(HarmonyLib.Transpilers.EmitDelegate<Func<int, int>>(miningId =>
```
Mushroom, line 33: Why is this required, is there a namespace collision now?

NebulaWorld/Logistics/ILSShipManager.cs
```diff
@@ -0,0 +1,304 @@
...
+                            break;
+                        }
+                    }
+                }
+                else
+                {
+                    int itemId = packet.itemId;
+                    int itemCount = packet.itemCount;
+                    stationComponent.TakeItem(ref itemId, ref itemCount);
+                    /*
```
Mushroom, line 231: Don't leave commented out code in

Mushroom (changes requested):
A few specific comments where I thought they were needed.

I dislike the lack of comments in this pull request, someone unfamiliar with this codebase (or this part of it, in my case) will have absolutely no idea what si going on a lot of the time any why certain things are done.

I would like to see some more comments that explain the code.

Mushroom (approved):

NebulaClient/PacketProcessors/Logistics/ILSAddStationComponentResponseProcessor.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace NebulaClient.PacketProcessors.Logistics
+{
+    [RegisterPacketProcessor]
+    public class ILSAddStationComponentResponseProcessor: IPacketProcessor<ILSAddStationComponentResponse>
+    {
+        public void ProcessPacket(ILSAddStationComponentResponse packet, NebulaConnection conn)
+        {
+            for(int i = 0; i < ILSShipManager.AddStationComponentQueue[packet.planetId].Count; i++)
+            {
+                if(ILSShipManager.AddStationComponentQueue[packet.planetId][i].shipDockPos == DataStructureExtensions.ToVector3(packet.shipDockPos))
```
Baldie-dev, line 23: Is it possible to put `DataStructureExtensions.ToVector3(packet.shipDockPos)` before the `for` loop? To prevent allocation of `new Vector3()` inside the loop.

NebulaHost/PacketProcessors/Logistics/ILSgStationPoolSyncRequestProcessor.cs
```diff
@@ -0,0 +1,155 @@
...
+ * Whenever a client connects we sync the current state of all ILS and ships to them
+ * resulting in a quite large packet but its only sent one time upon client connect.
+ */
+namespace NebulaHost.PacketProcessors.Logistics
+{
+    [RegisterPacketProcessor]
+    public class ILSgStationPoolSyncRequestProcessor : IPacketProcessor<ILSRequestgStationPoolSync>
+    {
+        private PlayerManager playerManager;
+        private const int maxShipsPerILS = 10;
```
Baldie-dev, line 17: Can this be unified with `const int ILSMaxShipCount = 10;` which is in the `NebulaClient.PacketProcessors.Logistics.ILSgStationPoolSyncProcessor`?

For example having this in both cases: `ILSShipManager.ILSMaxShipCount`

NebulaHost/PacketProcessors/Logistics/StationUIInitialSyncRequestProcessor.cs
```diff
@@ -0,0 +1,74 @@
...
+                }
+                if(stationComponent != null)
+                {
+                    StationStore[] storage = stationComponent.storage;
+                    int[] itemId = new int[storage.Length];
+                    int[] itemCountMax = new int[storage.Length];
+                    int[] itemCount = new int[storage.Length];
+                    int[] localLogic = new int[storage.Length];
+                    int[] remoteLogic = new int[storage.Length];
+                    int[] remoteOrder = new int[storage.Length];
```
Baldie-dev, line 42: I think there is space for future improvement, to have these arrays pre-allocated and just zero it here. Or maybe it does not need to be cleaned, since all values are being replaced right? So maybe just reusing the same arrays, so we would prevent allocation on every packet.

NebulaModel/Packets/Logistics/ILSShipDock.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.shipPos = new Double3[shipPos.Length];
+            this.shipRot = new Float4[shipRot.Length];
+            this.shipPPosTemp = new Double3[shipPPosTemp.Length];
+            this.shipPRotTemp = new Float4[shipPRotTemp.Length];
+
+            this.stationGId = stationGId;
+            this.shipDockPos = new Float3(shipDockPos);
+            this.shipDockRot = new Float4(shipDockRot);
+
+            Array.Copy(shipOtherGId, this.shipOtherGId, shipOtherGId.Length);
```
Baldie-dev, line 36: Why the copy is needed here? Would there be an issue if the whole array will be passed?
```cs
 this.shipOtherGId = shipOtherGId;
 ```
 Since this packet is being sent and the array is not modified, right?

File: NebulaModel/Packets/Logistics/ILSgStationPoolSync.cs [hunk 28]
```diff
@@ -0,0 +1,100 @@
...
+                                    int[] shipIndex,
+                                    Double3[] shipPos,
+                                    Float4[] shipRot,
+                                    Float3[] shipVel,
+                                    float[] shipSpeed,
+                                    Float3[] shipAngularVel,
+                                    Double3[] shipPPosTemp,
+                                    Float4[] shipPRotTemp)
+        {
+            this.stationGId = (int[])stationGId.Clone();
```
Baldie-dev, line 70: Do we need to clone it here? Will, there be any issue with just:
```cs
this.stationGId = stationGId;
```

NebulaModel/Packets/Logistics/StationUIInitialSync.cs
```diff
@@ -0,0 +1,70 @@
...
+                                    long energy,
+                                    long energyPerTick,
+                                    int[] itemId,
+                                    int[] itemCountMax,
+                                    int[] itemCount,
+                                    int[] localLogic,
+                                    int[] remoteLogic,
+                                    int[] remoteOrder)
+        {
+            this.itemId = new int[itemId.Length];
```
Baldie-dev, line 43: Same with before, is it possible to just assign the reference?
```cs
this.itemId = itemId;
```

NebulaPatcher/NebulaPlugin.cs
```diff
@@ -4,6 +4,7 @@
 using NebulaPatcher.Logger;
 using NebulaPatcher.MonoBehaviours;
 using System;
+using System.IO;
```
Baldie-dev, line 7: I think this dependency is not used, right?

Baldie-dev (approved):

PR #90, opened 2021-03-29
Storing and Loading player data (location + rotation)

Baldie-dev:
I have added following:

- Clients are now generating / loading keypair that is being sent (only public key) during the handshake
- Server creates SHA1 hash of the public key which is used as a unique identifier of the client
- This unique identifier is used as key in the `ThreadSafeDictionary<string, PlayerData> savedPlayerData`
- TODO: in the future, we can add some challenge-response authentication, but so at this moment only a public key is used.
- Server loads or creates `PlayerData` and sent it to the client.
- Client uses the location and rotation from it.

Video on [youtube](https://youtu.be/XlQIEUJvmNw).


NebulaModel/Utils/CryptoUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static string KeyFile = GameConfig.gameDocumentFolder + "player.key";
+
+        public static RSA GetOrCreateUserCert()
+        {
+            RSA rsa = RSA.Create();
+            if (rsa != null)
+            {
+                return rsa;
+            }
+            rsa = RSA.Create();
```
sp00ktober, line 20: Why do you do it 2 times?

Baldie-dev: I just removed it :D

NebulaWorld/SimulatedWorld.cs
```diff
@@ -227,10 +227,17 @@ public static void OnGameLoadCompleted()
...
-
+            // Change player location from spawn to the last known
+            VectorLF3 UPosition = new VectorLF3(LocalPlayer.Data.UPosition.x, LocalPlayer.Data.UPosition.y, LocalPlayer.Data.UPosition.z);
+            if (UPosition != VectorLF3.zero)
+            {
+                GameMain.mainPlayer.planetId = LocalPlayer.Data.LocalPlanetId;
+                GameMain.mainPlayer.position = LocalPlayer.Data.LocalPlanetPosition.ToUnity(); ;
+                GameMain.mainPlayer.uPosition = UPosition;
+                GameMain.mainPlayer.uRotation = Quaternion.Euler(LocalPlayer.Data.Rotation.ToUnity());
+                GameMain.mainPlayer.transform.localScale = Vector3.one;
```
sp00ktober, line 238: did you test this when spawning on a remote planet? Afaik you ened to set the `localScale` after the player was set as child to the planet (iirc this is what the game does when you are on a planet) which usually is after `ArrivePlanet()`, else you could end up with weird icarus sizes.

Baldie-dev: You are maybe right, I am checking it now. There is also some issue that if I leave the planet and come back I will be always spawned in the air. Hmm...

sp00ktober: maybe you need to recalculate the position based on planet movement, something like [this](https://github.com/sp00ktober/DSP_Mod/blob/2becb2f4abf7d280db6c935d1c65e499c6bed676/RenderDistance/Class1.cs#L230)

Baldie-dev: Go to 0:25s [video](https://www.youtube.com/watch?v=Lzpml9i8a24).
Seems like the player's transform parent is not the planet. Still looking into it.

sp00ktober: i think this is because you are not setting `localPlanet` (look at the globe in the corner it tells the wrong planet).
Thats why you start floating as the game does not think you are on that planet right now, it does not even load the `FactoryData` if im seeing this right.

Baldie-dev: right!

Baldie-dev: adding ` GameMain.data.ArrivePlanet(GameMain.data.galaxy.PlanetById(LocalPlayer.Data.LocalPlanetId));` seems to solve it. Test cases in [video](https://youtu.be/XlQIEUJvmNw):

- Load player data on the same planet
- Load player data from the space
- Load player data from another planet

Baldie-dev: oh wait, I will have to change it to `ArriveStar()` :(

NebulaWorld/SimulatedWorld.cs
```diff
@@ -227,10 +227,17 @@ public static void OnGameLoadCompleted()
...
+            // Change player location from spawn to the last known
+            VectorLF3 UPosition = new VectorLF3(LocalPlayer.Data.UPosition.x, LocalPlayer.Data.UPosition.y, LocalPlayer.Data.UPosition.z);
+            if (UPosition != VectorLF3.zero)
+            {
+                GameMain.data.ArrivePlanet(GameMain.data.galaxy.PlanetById(LocalPlayer.Data.LocalPlanetId));
+                GameMain.mainPlayer.planetId = LocalPlayer.Data.LocalPlanetId;
+                GameMain.mainPlayer.position = LocalPlayer.Data.LocalPlanetPosition.ToUnity(); ;
+                GameMain.mainPlayer.uPosition = UPosition;
+                GameMain.mainPlayer.uRotation = Quaternion.Euler(LocalPlayer.Data.Rotation.ToUnity());
+            }
```
hubastard, line 239: This seems wrong, you are always calling `ArrivePlanet()` even if the player is in space, shouldn't you guard it with a check to the `LocalPlanetId`?
Also isn't the `UPosition` always greater then 0,0,0 since it is the player Position in the Universe.

Baldie-dev: Now I am thinking about it and should move all of this to the handshake instead of having it in `OnGameLoadCompleted()`. The issue is, that client is joining the server with ` DSPGame.StartGameSkipPrologue(gameDesc);` after handshake is received. Which automatically spawns the player at the landing site by default. So the client will request the factory of the first planet and after everything is loaded it will move him to a different planet and request again the next factory. 
So I will move this code to the handshake so the proper factory is requested at first.
 
To your points:

- You are right about the `ArrivePlanet()`, I left it there is since there is already a check inside for this. But I will add the extra check so it will look better.
- In this case, `UPosition` is the latest position of the player and it is received from the server. So it will be always 0,0,0 when player connects for the first time. Client automatically spawns player at the spawn position based on the `GameDesc` data, so there is no need to change player position in that case.

Baldie-dev:
In the last commits I have added:

- `GameData_Patch.cs` -> `[HarmonyPatch("SetForNewGame")]`
Which made the client request only star and planet where he currently is.

- `OnGameLoadCompleted `-> `OnGameLoadCompleted()`  
 The Player's location is calculated based on the current position of the planet. This is a bug fix for the issue where the planet moved away.

NebulaPatcher/Patches/Dynamic/GameData_Patch.cs
```diff
@@ -114,6 +114,30 @@ public static bool OnActivePlanetFactoryLoaded_Prefix(GameData __instance, Plane
             return false;
         }
 
+        [HarmonyPostfix]
+        [HarmonyPatch("SetForNewGame")]
+        public static void Postfix(GameData __instance)
```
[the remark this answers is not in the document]
Baldie-dev: sorry, miss that. Fixed.

hubastard (approved):

Issue #96, opened 2021-04-02
Saving player's data


Baldie-dev:
I am creating this for tracking purposes since I have started to work on it.

The following needs to be finished:

- [x] Every player will generate key pair for identification.
- [x] Host will store the last position and color of the player.
- [x] Host will store player's inventory.
- [x] Host will store other info such as (mecha energy, sand count, etc.)
- [x] Host will create a separate `.server` save-file with info about all players.

PR #97, opened 2021-04-03
Saving Player's Mecha's data (Inventory, Energy, Replicator, etc.)

Baldie-dev:
This is part of the #96 .

Following changes were added:

- I have added the data structure `MechaData` that contains all info about the Mech.
- This object is part of the `PlayerData` on the host, so the server keeps a track of it.
- Player is regularly sending `MechaData` updates to the host to save it.
- When a player joins the server, the host will load `MechaData` from memory + it will attach current tech bonuses and it will send it to the client as part of the `PlayerData`.
- Client will use it to update its own mech.

Video on [Youtube](https://youtu.be/50GUinb2jgI).

Changes to current code:

- I am creating a copy of `PlayerData` object without ` MechaData`  when other players are notified that the player is joining since they do not need to know about the player's inventory.
```
PlayerData pdata = player.Data.CreateCopyWithoutMechaData(); // Remove inventory from mecha data
foreach (Player activePlayer in playerManager.GetConnectedPlayers()) {
                activePlayer.SendPacket(new PlayerJoining(pdata)); }
```

### DIFF
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -59,7 +59,7 @@ namespace NebulaClient
         void Disconnect()
         {
             IsConnected = false;
-            clientSocket.Close((ushort)NebulaStatusCode.ClientRequestedDisconnect, "Player left the game");
+            clientSocket.Close((ushort)DisconnectionReason.ClientRequestedDisconnect, "Player left the game");
         }
 
         public void DestroySession()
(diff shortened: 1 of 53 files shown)


Baldie-dev:
In the last commit, the server will be storing PlayerData in the extra save-file.
video on [youtube](https://www.youtube.com/watch?v=uvUoq71PCJw).

hubastard (approved):

Issue #101, opened 2021-04-04
Storage System Synchronization


Baldie-dev:
I am creating this for tracking purposes since I have started to work on it.

To Do:

- [x] Initial Storage sync, when player click on the storage
- [x] Sending real-time updates of the storage that is currently opened for the user.
- [x] synchronize icons of the storages
- [x] synchronize events from other players

PR #103, opened 2021-04-06
Storage System Synchronization

Baldie-dev:
This is part of the #101 .

The following has been synchronized:

- When a user clicks on the storage, the loading message is displayed in the storage window. Once the server replies with the storage data, the loading message is replaced with the actual inventory.
  img
- If any user or host makes some storage transfer, the event is broadcasted to everyone on the same planet. 
- Storage inventory sorting
- Setting storage filter (maximum amount of items that can inserter insert.)

Video on [youtube](https://youtu.be/Vat8kwPUXrI).


Baldie-dev:
conflict resolved :)

hubastard (approved):

Issue #104, opened 2021-04-06
Interaction with Buildings Synchronization


Baldie-dev:
I am creating this for tracking purposes since I have started to work on it.

ToDo:

- [x] Assembler interactions
- [x] Furnace interactions
- [x] Research station interactions
- [x] Rafinery interactions
- [x] Particle collider interactions
- [x] Railgun interactions
- [x] Launching silo interactions
- [x] Receiver interactions
- [x] Splitter interactions
- [ ] Power Generator interactions
- [x] Inserter interactions
- [x] Miner interactions
- [x] Fluid storage interactions
- [x] Power Exchanger interactions

PR #106, opened 2021-04-07
Assembler Interaction Sync + Planet broadcast

Baldie-dev:
This is part of the #104 

This PR contains two parts:
**1. Assembler interactions**
- Reset recipe, set recipe, add and withdraw items
- Video on [youtube](https://youtu.be/ClrPg7kc7dE)

**2. Broadcasting packets to players on the particular planet**
  - This was added to have an easy way how to send a packet only to the players that are on the same planet as the sender, via `LocalPlayer.SendPacketToLocalPlanet()` as used in `UIAssemblerWindow_Patch.cs`.
  - From the host side:
    - Host will just simply go through the list of players and send packets to the players that are on the same planet as the host.
  - From the client-side:
    - Client will insert the packet inside the `PlanetBroadcastPacket`, adds the player's planet id, and send it to the host.
    - Host will process it in `PlanetBroadcastProcessor` by taking the planet id and forwarding the inserted packet to every user that is on the same planet. If the host is on the target planet, it will add the packet to his own queue for processing.

  - I test this with 3 players and it seems to be working fine, video on [youtube](https://www.youtube.com/watch?v=--1S4k12gV0).


NebulaHost/PlayerManager.cs
```diff
@@ -70,6 +70,28 @@ public void SendPacketToAllPlayers<T>(T packet) where T : class, new()
             }
         }
 
+        public void SendPacketToLocalPlanet<T>(T packet) where T : class, new()
+        {
+            foreach (Player player in GetConnectedPlayers())
+            {
+                if (player.Data?.LocalPlanetId == GameMain.localPlanet?.id)
```
hubastard, line 77: Maybe I'm wrong, but I think that right now we might not update the `LocalPlanetId` in the `PlayerData` after the initial game load.

Baldie-dev: yeah, I think it is 0 for the new `PlayerData` and it gets updated at the first `PlayerMovement` packet. For the players, that are reconnecting, it should be already populated from the previous saved state.

I guess I can set the `LocalPlanetId` to the spawn planet for the new players.

hubastard (approved):

PR #108, opened 2021-04-08
Add player names to the starmap and in-game

Mushroom:
This adds the location of other players to the star map

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/UIStarmap_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/UIStarmap_Patch.cs
@@ -1,9 +1,5 @@
 ﻿using HarmonyLib;
 using NebulaWorld;
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 
 namespace NebulaPatcher.Patches.Dynamic
 {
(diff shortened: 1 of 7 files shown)


NebulaPatcher/Patches/Dynamic/UIStarmap_Patch.cs
```diff
@@ -0,0 +1,27 @@
+﻿using HarmonyLib;
+using NebulaWorld;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
```
Baldie-dev, line 6: These dependencies are not used here, right?

Mushroom: Done

Baldie-dev (approved):

NebulaPatcher/Patches/Dynamic/UIStarmap_Patch.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            SimulatedWorld.RenderPlayerNameTagsOnStarmap(__instance);
+        }
+
+        [HarmonyPostfix]
+        [HarmonyPatch("_OnClose")]
+        public static void OnClose_Postfix(UIStarmap __instance)
+        {
+            SimulatedWorld.ClearPlayerNameTagsOnStarmap();
+        }
```
hubastard, line 21: Shouldn't this be skipped if not in multiplayer session ?

Mushroom: Nothing would happen as there would not be any remote players in the list, but I can add a check

PR #109, opened 2021-04-08
Prebuilds Sync

Baldie-dev:
I have changed how Prebuilds is being synchronized:

1. When players click during building, he will send a list of `buildPreviews` to the host and the host will send it to all players.
2. I have to manually serialize and deserialize the `buildPreview` object.
3. Once the client receives the `buildPreviews`, it will call `CreatePrebuilds();` which will calculates all connections and final state/shape of the building.

Note: I have to add a transpiler to prevent taking items from the inventory if the player is not a builder but only a receiver.

### DIFF
--- a/NebulaHost/PacketProcessors/Factory/CreatePrebuildsRequestProcessor.cs
+++ b/NebulaHost/PacketProcessors/Factory/CreatePrebuildsRequestProcessor.cs
@@ -1,0 +1,42 @@
+﻿using NebulaModel.Attributes;
+using NebulaModel.Networking;
+using NebulaModel.Packets.Factory;
+using NebulaModel.Packets.Processors;
+using NebulaWorld.Factory;
+using HarmonyLib;
+using System.Collections.Generic;
+
+namespace NebulaHost.PacketProcessors.Factory
+{
+    [RegisterPacketProcessor]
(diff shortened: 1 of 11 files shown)


NebulaHost/PacketProcessors/Factory/CreatePrebuildsRequestProcessor.cs
```diff
@@ -0,0 +1,42 @@
...
+                //Make backup of values that are overwritten
+                List<BuildPreview> tmpList = pab.buildPreviews;
+                bool tmpConfirm = pab.waitConfirm;
+                UnityEngine.Vector3 tmpPos = pab.previewPose.position;
+                UnityEngine.Quaternion tmpRot = pab.previewPose.rotation;
+
+                //Create Prebuilds from incomming packet
+                pab.buildPreviews = packet.GetBuildPreviews();
+                pab.waitConfirm = true;
+                FactoryManager.EventFromServer = true;
```
hubastard, line 27: This should probably be an `FactoryManager.EventFromClient` since we are in the host code here.

hubastard: I'll fix it in main

Issue #132, opened 2021-04-11
All players will receives items in storage of building when it is destroyed.


Baldie-dev:
I am already making fix for this:
![image](https://user-images.githubusercontent.com/79051050/114303158-f804c480-9acc-11eb-9c2a-44e5993f3c07.png)

Baldie-dev:
retested and does not happens now.

Issue #127, opened 2021-04-10
Clients are able to duplicate buildings by destroying their prebuild with a double click.


hubastard:
At first I had 7 labs and by double clicking the prebuild while in destroy mode, it calls the destroy function twice on the client and duplicated the item.

https://user-images.githubusercontent.com/69216913/114275291-f6b19a00-99ef-11eb-8e2c-689989074048.mp4

Baldie-dev:
Bug fix in progress :)

Baldie-dev:
retested and does not happens now.

Issue #115, opened 2021-04-09
Clients were able to stack research labs above the current stack limit.


hubastard:
During our playthrough, a client was able to stack 6 labs even if the the limit was 3.

hubastard:
@sp00ktober We would need to test it again while being 3 players, I think that it was related to the desync issue that I fixed https://github.com/hubastard/nebula/commit/aeb063e8b0071fc4a977441432edabbb39c5b7bf

hubastard:
Ok no, the issue is still there. I've been able to replicate it but only on a client not the host.

Baldie-dev:
I am checking it now :)

Baldie-dev:
Hmm, i have ordered additional RAM :D 3rd instance is crashing with out of memory errors.
But I am not able to reproduce it with client and host alone :( the stack level 3 should be base without any research techs unlocked. I am not sure how it can increase on its own :(

Baldie-dev:
It is constant "labLevel" in GameHistoryData object, that is being synced by import/export functionality. Hmmm

hubastard:
In my test it took a maybe a minute of building / destroying before I was able to reproduce the issue. When the issue occurred, I was building 2 stacks at the same time alternating the building from one tower to the other. Once it "broke" it was causing the issue 100% of the time. But it took some time before

hubastard:
I'll try to reproduce the issue and make a video, maybe its a vanilla bug

Baldie-dev:
I think there will be some external factor to it. I was trying multiple combinations of destroying and building new levels from different clients and host, but no luck with reproducing it. I will check it from the code point of view, how this can happens:
![image](https://user-images.githubusercontent.com/79051050/114263953-7cc9e280-99e8-11eb-9352-6b8730977955.png)

hubastard:
Was finally able to replicate this on video:
https://youtu.be/l1UG7pdzkKY

Baldie-dev:
This should be fixed now :) After we do another test, we can close it.

Issue #94, opened 2021-04-01
sinking under the terrain bug


sp00ktober:
https://user-images.githubusercontent.com/79335616/113276673-5c12d600-92e0-11eb-941d-465b7d5b3b73.mp4

Steps i did beofre this happened:
- client travels to the planet, then leaves again into space. (this way ``FactoryData`` is received and stored locally)
- host travels there, places building (not sure if its only triggered by ILS)
- client comes back, tadaa

This _may_ be caused by ``FlattenTerrain()`` which ic called when the building is placed.

hubastard:
So this issue occurs because when we place a building the code calls a FlattenTerrain, but since we don't have the distant planet mesh that is being loaded, the ground collision that is generated in the FlattenTerrain is just an empty mesh (no collider at all)

hubastard:
Right now, I can think of 2 ways of fixing this:
- We can try to find a way to call or replicate what the `PlanetModelingManager.RequestLoadPlanetFactory()` does to generate the planet mesh.
- We could also store all the FlattenTerrain operation that could not be perform on the host or on remote clients and request the list of FlattenTerrain operation that we have missed when we load the planet. We should also either save those operation to disk on the host or apply them all before closing the host game.

If anyone have a better idea please share your ideas 😄

Baldie-dev:
Retested and seems to be closed

aschmois:
By the way, I'm still seeing this exact problem pop up only for the host machine. I'm currently on EA 0.9.25.12201 Nebula 0.8.8.0, no extra mods installed. A restart to the main menu and back fixes the issue. I believe this problem happens when changing planets only.

PR #134, opened 2021-04-12
Bug Fixes + Prebuild Enhancement

Baldie-dev:
This PR contains bug fixes for:

- #132 **(All players will receive items in the storage of the building when it is destroyed.)**
  - I have added `AuthorId` value to the destroy request packet. Once the server confirms that the building is destroyed, the author will get its items.
  
- #127 **(Clients are able to duplicate buildings by destroying their prebuild with a double click.)**
  - Client is not getting the building item immediately upon click on the building. They have to wait for the confirmation packet from the server
 
- #115 **(Clients were able to stack research labs above the current stack limit.)**
  - Added a new implementation of validation if the client can build the building. The server test's the building placement by calling the original method `CheckBuildConditions()` and then it decides if the building packet should be confirmed or not.

- #94 **(sinking under the terrain bug)**
  - Added null check for mesh colliders when player arrives to planet:
```cs
 if (planetData.meshColliders[i].sharedMesh == null)
  {
    planetData.meshColliders[i].sharedMesh = planetData.meshes[i];
  }
```

### DIFF
--- a/NebulaClient/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
+++ b/NebulaClient/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
@@ -3,6 +3,7 @@ using NebulaModel.Attributes;
 using NebulaModel.Networking;
 using NebulaModel.Packets.Factory;
 using NebulaModel.Packets.Processors;
+using NebulaWorld;
 using NebulaWorld.Factory;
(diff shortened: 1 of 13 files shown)


Baldie-dev:
Since this PR also transfers more data about the BuildPreview, it also fixes the:
#136 **(Clients cannot continuously build belt path)**
output

hubastard (approved):

Issue #126, opened 2021-04-10
Player's drone synchronization


Baldie-dev:
I am opening this issue for tracking purposes since I have started to work on it.

Synchronization of the drones that are building stuff around the players, so that is also visible to other players.

PR #145, opened 2021-04-15
Player's drone sync + labor division

Baldie-dev:
This is the implementation of #126 

I am basically sending 3 events using transpilers:

- Drone went to build something
- Drone is going back to mecha
- Drone changed build target

The labor division (each drone will build something else) is achieved by leveraging the original approach of marking which prebuild is under build from the drone. In case of collision (client and host will send drone at the same time), a random number is generated as a priority for the drone order. 

For example:
1. Host will sent packet: my drone is going to build that with the random number `178361`.
2. Client will sent packet: my drone is going to build that with the random number `2781`.

In this case, the Client lost and his drone has to go back to mecha. In the case of more players, the higher number will always win. 

In the GIF below, I have build 8 buildings and each player has only 3 drones. That is why the first 6 buildings are built quickly and the last two later.

output

### DIFF
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -81,6 +81,10 @@ namespace NebulaClient
         {
             serverConnection?.SendPacket(new StarBroadcastPacket(PacketProcessor.Write(packet), GameMain.data.localStar?.id ?? -1));
         }
+        public void SendPacketToLocalPlanet<T>(T packet) where T : class, new()
+        {
+            serverConnection?.SendPacket(new PlanetBroadcastPacket(PacketProcessor.Write(packet), GameMain.mainPlayer.planetId));
+        }
 
         public void Reconnect()
         {
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -85,6 +85,11 @@ namespace NebulaClient
         {
             serverConnection?.SendPacket(new PlanetBroadcastPacket(PacketProcessor.Write(packet), GameMain.mainPlayer.planetId));
         }
+        public void SendPacketToPlanet<T>(T packet, int planetId) where T : class, new()
+        {
+            //Should send packet to particular planet
+            //Not needed at the moment, used only on the host side
+        }
 
         public void Reconnect()
         {
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -6,6 +6,7 @@ using NebulaModel.Packets.Routers;
 using NebulaModel.Packets.Session;
 using NebulaModel.Utils;
 using NebulaWorld;
+using System;
 using System.Net;
 using UnityEngine;
 using WebSocketSharp;
@@ -89,6 +90,7 @@ namespace NebulaClient
         {
             //Should send packet to particular planet
             //Not needed at the moment, used only on the host side
+            throw new NotImplementedException();
         }
 
         public void Reconnect()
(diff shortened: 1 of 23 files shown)


Baldie-dev:
I will resolve conflicts + two bug fixes are needed... please do not merge yet

Baldie-dev:
It should be ready for the merge now :) all issues solved.

NebulaWorld/SimulatedWorld.cs
```diff
@@ -286,25 +353,28 @@ public static void OnDronesDraw()
...
-            double tmp2 = 0;
+            double tmp = 1e10; //fake energy of remote player, needed to do the Update()
+            double tmp2 = 1;
             //Update drones positions based on their targets
             foreach (KeyValuePair<ushort, RemotePlayerModel> remoteModel in remotePlayersModels)
             {
                 MechaDrone[] drones = remoteModel.Value.PlayerInstance.mecha.drones;
-                int droneCount = remoteModel.Value.PlayerInstance.mecha.droneCount;
+                int droneCount = GameMain.mainPlayer.mecha.droneCount;
                 for (int i = 0; i < droneCount; i++)
```
hubastard, line 373: This feels wrong to me, you iterate over the remote players drones by using the local player drones count ??

Shad0wlife: Well is the drones count the count of available or active drones? If available, the tech level is synced, so the count should be the same.

NebulaClient/MultiplayerClientSession.cs
```diff
@@ -81,6 +81,15 @@ public void SendPacketToLocalStar<T>(T packet) where T : class, new()
...
             serverConnection?.SendPacket(new StarBroadcastPacket(PacketProcessor.Write(packet), GameMain.data.localStar?.id ?? -1));
         }
+        public void SendPacketToLocalPlanet<T>(T packet) where T : class, new()
+        {
+            serverConnection?.SendPacket(new PlanetBroadcastPacket(PacketProcessor.Write(packet), GameMain.mainPlayer.planetId));
+        }
+        public void SendPacketToPlanet<T>(T packet, int planetId) where T : class, new()
+        {
+            //Should send packet to particular planet
+            //Not needed at the moment, used only on the host side
```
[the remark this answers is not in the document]
Baldie-dev: yup, I agree. I will add it

NebulaWorld/SimulatedWorld.cs
Baldie-dev: It seemed to be unnecessary to sync also `droneCount` for every remote model since this value is the same for all players. This number represents how many drones the mecha has which is the same for all players.
If I would iterate over the remote player's drones count I would also have to implement synching of a number of drones for remote mechas everything someone joins / every time some research unlocks it. And for me, it was an unnecessary step.

hubastard: Ok yeah that's good then.

hubastard (approved):

Baldie-dev:
ready for merge

NebulaHost/PacketProcessors/Routers/PlanetBroadcastProcessor.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            playerManager = MultiplayerHostSession.Instance.PlayerManager;
+        }
+        public void ProcessPacket(PlanetBroadcastPacket packet, NebulaConnection conn)
+        {
+            Player player = playerManager.GetPlayer(conn);
+            if (player != null)
+            {
+                //Forward packet to other users
+                playerManager.SendRawPacketToPlanet(packet.PacketObject, packet.PlanetId, conn);
```
Mushroom, line 22: Wouldn't it be better to include the host in this call rather than individually below?

Mushroom (changes requested):

NebulaHost/PacketProcessors/Routers/PlanetBroadcastProcessor.cs
Baldie-dev: you are right, I should do the check for the same planet in the specific packet processor and not here. I will fix

Mushroom (approved):

Issue #146, opened 2021-04-15
Foundation building synchronization


Baldie-dev:
I am creating this for tracking purposes since I have started to work on it.

PR #149, opened 2021-04-15
Foundation Sync + Star Broadcast fix

Baldie-dev:
This is the implementation of #146 

I have to also add a fix for the broadcasting of star changes.

### DIFF
--- a/NebulaClient/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
+++ b/NebulaClient/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
@@ -1,0 +1,60 @@
+﻿using NebulaModel.Attributes;
+using NebulaModel.DataStructures;
+using NebulaModel.Networking;
+using NebulaModel.Packets.Factory;
+using NebulaModel.Packets.Processors;
+using NebulaWorld.Factory;
+using UnityEngine;
+
+namespace NebulaClient.PacketProcessors.Factory.Foundation
+{
+    [RegisterPacketProcessor]
+    class FoundationBuildUpdateProcessor : IPacketProcessor<FoundationBuildUpdatePacket>
+    {
+        public void ProcessPacket(FoundationBuildUpdatePacket packet, NebulaConnection conn)
+        {
+            //Check if client has loaded planet
+            PlanetData planet = GameMain.galaxy.PlanetById(packet.PlanetId);
+            PlanetFactory factory = planet?.factory;
+            if (factory != null)
+            {
+                Vector3[] reformPoints = new Vector3[100];
+                Vector3 reformCenterPoint = new Vector3();
+                if (factory.platformSystem.reformData == null)
+                {
+                    factory.platformSystem.InitReformData();
+                }
+                if (planet.physics == null || planet.physics.colChunks == null)
+                {
+                    planet.physics = new PlanetPhysics(planet);
+                    planet.physics.Init();
+                }
+                if (planet.aux == null)
+                {
+                    planet.aux = new PlanetAuxData(planet); 
+                }
+                int reformPointsCount = factory.planet.aux.ReformSnap(packet.GroundTestPos.ToVector3(), packet.ReformSize, packet.ReformType, packet.ReformColor, reformPoints, packet.ReformIndices, factory.platformSystem, out reformCenterPoint);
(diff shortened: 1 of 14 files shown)


hubastard (approved):

NebulaHost/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+using NebulaModel.Networking;
+using NebulaModel.Packets.Factory;
+using NebulaModel.Packets.Processors;
+using NebulaWorld.Factory;
+using UnityEngine;
+
+namespace NebulaHost.PacketProcessors.Factory.Foundation
+{
+    [RegisterPacketProcessor]
+    class FoundationBuildUpdateProcessor : IPacketProcessor<FoundationBuildUpdatePacket>
```
Therzok, line 12: Are these two files identical? Why not use linked files?

Baldie-dev: I think there are more cases where PacketProcessor files are completely the same on the client and on the host. I guess we can use linked files, I just have to check, how it is done.
Also, I think the namespace needs to be:
for host - namespace NebulaHost.*
for client- namespace NebulaClient.*

I guess. But that is good note.

NebulaClient/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
```diff
@@ -0,0 +1,60 @@
...
+    class FoundationBuildUpdateProcessor : IPacketProcessor<FoundationBuildUpdatePacket>
+    {
+        public void ProcessPacket(FoundationBuildUpdatePacket packet, NebulaConnection conn)
+        {
+            //Check if client has loaded planet
+            PlanetData planet = GameMain.galaxy.PlanetById(packet.PlanetId);
+            PlanetFactory factory = planet?.factory;
+            if (factory != null)
+            {
+                Vector3[] reformPoints = new Vector3[100];
```
Therzok, line 21: Looks like something that could be cached for the processor, since there's no concurrent users.

NebulaHost/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
Therzok: There is a discussion on Discord about this, @hubastard wants to think on it first.

NebulaHost/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public void ProcessPacket(FoundationBuildUpdatePacket packet, NebulaConnection conn)
+        {
+            PlanetData planet = GameMain.galaxy.PlanetById(packet.PlanetId);
+            PlanetFactory factory = GameMain.data.GetOrCreateFactory(planet);
+            Vector3[] reformPoints = new Vector3[100];
```
Therzok, line 18: Looks like a reusable buffer.

NebulaClient/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
Baldie-dev: This is the "output" variable that will have the result from this method
```cs
ComputeFlattenTerrainReform(reformPoints, ...
```

But I guess I can cache the allocated array so it will not be allocated each time the packet is received and just clean the array with empty values.

NebulaHost/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
Baldie-dev: I will change it :)

NebulaClient/PacketProcessors/Factory/Foundation/FoundationBuildUpdateProcessor.cs
Therzok: `ArrayPool<T>` is only available in newer target frameworks, sadly. This is not a big issue btw, it's just that it might cause gen0 GCs in done frequently.

Issue #165, opened 2021-04-17
Client will get exception, if host mines vegetables or remote planets


Baldie-dev:
I am just fixing this:
```cs
InvalidOperationException: Nullable object must have a value.
  at System.Nullable`1[VegeData].get_Value () [0x00000] in <filename unknown>:0 
  at NebulaWorld.SimulatedWorld.MineVegetable (NebulaModel.Packets.Planet.VegeMined packet) [0x00000] in <filename unknown>:0 
  at NebulaClient.PacketProcessors.Planet.VegetationMinedProcessor.ProcessPacket (NebulaModel.Packets.Planet.VegeMined packet, NebulaModel.Networking.NebulaConnection conn) [0x00000] in <filename unknown>:0 
  at NebulaModel.Networking.Serialization.NetPacketProcessor+<>c__DisplayClass34_0`2[NebulaModel.Packets.Planet.VegeMined,NebulaModel.Networking.NebulaConnection].<SubscribeReusable>b__0 (NebulaModel.Networking.Serialization.NetDataReader reader, System.Object userData) [0x00000] in <filename unknown>:0 
  at NebulaModel.Networking.Serialization.NetPacketProcessor.ReadPacket (NebulaModel.Networking.Serialization.NetDataReader reader, System.Object userData) [0x00000] in <filename unknown>:0 
  at NebulaModel.Networking.Serialization.NetPacketProcessor.ProcessPacketQueue () [0x00000] in <filename unknown>:0 
  at NebulaClient.MultiplayerClientSession.Update () [0x00000] in <filename unknown>:0 
 ```

Issue #164, opened 2021-04-16
Vegetables needs to be mined twice on remote planets


Baldie-dev:
On the remote planet, players need to mine vegetables twice in order to destroy it. They are probably generated twice.

[youtube](https://www.youtube.com/watch?v=rabfVCWFIH8)

Baldie-dev:
I am on it

UeliDeSchwert:
Just a note: You sure you mean "Vegetables", not "Vegetation"? 😆

Issue #139, opened 2021-04-13
Vege are not destroyed if player is not on the planet


Baldie-dev:
On FlattenTerrain() the vege objects like rocks, trees, etc. should be destroyed. But if player is not on the planet, they are not loaded and therefore not destroyed :frowning: 

To find which vege should be destroyed, they use `nearColliderLogic.GetVegetablesInAreaNonAlloc()` to go though all colliders and find those who collide with the building. And I really really do not want to load the whole planet just to place one building.

So I will make some queue and once player visits the planet the vege in queue will be destroyed.

Baldie-dev:
I will create separate PR for this.

Baldie-dev:
![image](https://user-images.githubusercontent.com/79051050/115090353-ea0fd300-9f14-11eb-96de-9d5afa80468b.png)

PR #166, opened 2021-04-17
Destroying vegetables on remote planets sync, preventing double vege generation + bug fix

Baldie-dev:
Following PR solves three issues:

1. #164 - Vegetables were generated twice for the remote planets.
2. #165 - Fix for the exception that was raised on the client-side when host mines on a remote planet.
3. #139 - Vegetables were not destroyed if the player was not on the planet and did not have loaded colliders (for example by calling "FlattenTerrain", that function needs colliders to detect which vegetables need to be destroyed)


sp00ktober (approved):
Looks good to me.

hubastard (approved):

Issue #160, opened 2021-04-16
Auto save failure


fudgey297:
This auto save failure message arrived as soon as we left the orbit of our first planet.
![image](https://user-images.githubusercontent.com/29759060/115056417-bb750680-9eda-11eb-85d8-93e9814f8056.png)

PR #167, opened 2021-04-17
Null check for updating colliders

Baldie-dev:
This is a fix for the #160 

image

I am not sure why, but this is caused by a race condition when ArrivePlanet and GameTick are called before the planet is loaded.

Added null check that will delay the refresh.


Baldie-dev:
These two lines were added:
image

Mushroom (approved):
LGTM

hubastard (approved):

Issue #169, opened 2021-04-17
Tech upgrades are not synced for the new players


Baldie-dev:
When a player that has never been on the server joins, he will not receive the tech upgrades.

I am currently working on it.

PR #170, opened 2021-04-17
Bug fix: Loading mecha upgrades for new player

Baldie-dev:
I had a small issue that mecha upgrades were loaded only for the players whose save state already exists but it did not work for new players.

This will fix issue #169


Baldie-dev:
Also added null check for the belt presence if someone put stuff on the belt and at the same time the belt is removed.

So this error is fixed:
image

hubastard (approved):

PR #179, opened 2021-04-17
Fix PrebuildsRequestProcessor not setting the right value.

Therzok:
In case one of the inputs was wrong, but the outputs were right, since the wrong variable was assigned, this caused the function to return the wrong value, so refactor it a bit

### DIFF
--- a/NebulaHost/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
+++ b/NebulaHost/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
@@ -79,7 +79,7 @@ namespace NebulaHost.PacketProcessors.Factory.Entity
                 GameMain.mainPlayer.mecha.buildArea = float.MaxValue;
                 FactoryManager.IgnoreBasicBuildConditionChecks = true;
                 bool canBuild = pab.CheckBuildConditions();
-                canBuild &= CheckBuildingConnections(pab.buildPreviews, planet.factory);
+                canBuild &= CheckBuildingConnections(pab.buildPreviews, planet.factory.entityPool, planet.factory.prebuildPool);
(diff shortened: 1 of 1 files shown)


NebulaHost/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
```diff
@@ -110,31 +110,44 @@ public void ProcessPacket(CreatePrebuildsRequest packet, NebulaConnection conn)
...
-                {
-                    isInputOk = factory.entityPool.Length >= buildPreviews[i].inputObjId && factory.entityPool[buildPreviews[i].inputObjId].id != 0;
-                } else if (buildPreviews[i].inputObjId < 0)
-                {
-                    isInputOk = factory.prebuildPool.Length >= -buildPreviews[i].inputObjId && factory.prebuildPool[-buildPreviews[i].inputObjId].id != 0;
-                }
-                bool isOutputOk = true;
-                if (buildPreviews[i].outputObjId > 0)
-                {
-                    isInputOk = factory.entityPool.Length >= buildPreviews[i].outputObjId && factory.entityPool[buildPreviews[i].outputObjId].id != 0;
```
Therzok, line 129: See here, isOutputOk should've been set. I've refactored it a bit, makes it easier to follow.

Especially since `factory.entityPool.Length == 3` means `0,1,2` are valid, but this allows `entityPool[3]` to be considered valid.

Baldie-dev: I agree, this looks much better.

Baldie-dev (approved):

Baldie-dev:
Looks much better than my approach :)

PR #184, opened 2021-04-17
Fix thread safe issue with RemotePlayerModels

hubastard:
Fix thread safe issue that was causing a null ref issue when a client disconnect.

### DIFF
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -261,7 +261,7 @@ namespace NebulaWorld
             int localPlanetId = -1;
             using (GetRemotePlayersModels(out var remotePlayersModels))
             {
-                if (!remotePlayersModels.TryGetValue((ushort)packet.PlayerId, out RemotePlayerModel pModel))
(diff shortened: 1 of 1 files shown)


NebulaWorld/SimulatedWorld.cs
```diff
@@ -82,151 +98,183 @@ public static void UpdateGameState(GameState state)
...
         }
 
         public static void MineVegetable(VegeMined packet)
         {
-            if (!remotePlayersModels.TryGetValue((ushort)packet.PlayerId, out RemotePlayerModel pModel))
+            int localPlanetId = -1;
+            using (GetRemotePlayersModels(out var remotePlayersModels))
             {
-                Debug.Log("FAILED TO SYNC VEGE DATA");
+                if (!remotePlayersModels.TryGetValue((ushort)packet.PlayerId, out RemotePlayerModel pModel))
```
Therzok, line 264:
```suggestion
                if (remotePlayersModels.TryGetValue((ushort)packet.PlayerId, out RemotePlayerModel pModel))
```

PR #185, opened 2021-04-18
Fix for Mecha Drones having infinite range

Jon-Myrstad:
Fixed an issue with mecha.buildArea permanently stays as float.MaxValue after PrebuildsRequestProcessor ProcessPacket is ran.


Baldie-dev (approved):

Shad0wlife:
I don't think this actually fixed the issue. We used the build from the check and the clients drones flew around a quarter planet to build some belts for me still. Can someone confirm?

Baldie-dev:
I think it is just too big drone distance. Will be adjusted in next PR :)

Jon-Myrstad:
Are you in a previously used save @Shad0wlife? Because the broken range gets stored in the save file.

Could restore the default range if it has been broken once, but I'm not sure if there's anything that is supposed to increase it, like upgrades, or if "80" is what it always is.

Shad0wlife:
Hey! Yeah that might explain it 😅

PR #186, opened 2021-04-18
Fixed an issue where gas planets throws a null ref exception because …

Jon-Myrstad:
…of a missing meshcollider at index 16. I assume this is because it has no actual terrain, but don't know for certain why.


Therzok (approved):

Baldie-dev:
weird, I thought that whole planet mesh is loaded at once. But this should be good. Thanks!

Baldie-dev (approved):

PR #189, opened 2021-04-18
Fixes sorting your own Mecha inventory in space as it was throwing a …

Jon-Myrstad:
…null ref exception.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/StorageComponent_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/StorageComponent_Patch.cs
@@ -15,7 +15,7 @@ namespace NebulaPatcher.Patches.Dynamic
         public static bool AddItem_Prefix(StorageComponent __instance, int itemId, int count, int startIndex, int length)
         {
             //Run only in MP, if it is not triggered remotly and if this event was triggered manually by an user
-            if (SimulatedWorld.Initialized && !StorageManager.EventFromServer && !StorageManager.EventFromClient && StorageManager.IsHumanInput)
+            if (SimulatedWorld.Initialized && !StorageManager.EventFromServer && !StorageManager.EventFromClient && StorageManager.IsHumanInput && GameMain.data.localPlanet != null)
             {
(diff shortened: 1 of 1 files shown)


Baldie-dev:
This sort function should be synchronized only for the storage buildings. So in the function `HandleUserInteraction` there is a check to ignore interaction with mecha's inventory.

You are right, but I would change the condition from:
```cs
&& GameMain.data.localPlanet != null
```
to:
```cs
&& __instance.entityId > 0 && __instance.id > 0
```

And remove check: `&& __instance.entityId > 0 && __instance.id > 0` from the `HandleUserInteraction` function below. The reason is, that we should ignore it for the mecha inventory, but using `&& GameMain.data.localPlanet != null` it would still process further, it would create a packet and after that, it would check for mecha inventory and then returns. So it would be unneccessary allocation.

Baldie-dev:
Could you please also add the null check to the function above `AddItem`, `AddItemStacked`, `TakeItemFromGrid `?
Thanks a lot!

NebulaPatcher/Patches/Dynamic/StorageComponent_Patch.cs
```diff
@@ -61,7 +61,7 @@ public static void SetBans_Postfix(StorageComponent __instance, int _bans)
         [HarmonyPatch("Sort")]
         public static void Sort_Postfix(StorageComponent __instance)
         {
-            if (SimulatedWorld.Initialized && !StorageManager.EventFromServer && !StorageManager.EventFromClient)
+            if (SimulatedWorld.Initialized && !StorageManager.EventFromServer && !StorageManager.EventFromClient && GameMain.data.localPlanet != null)
```
Baldie-dev, line 64: Could you please also add the null check to the functions above `AddItem`, `AddItemStacked`, `TakeItemFromGrid`?
Thanks a lot!

Baldie-dev (changes requested):

NebulaPatcher/Patches/Dynamic/StorageComponent_Patch.cs
Jon-Myrstad: Didn't do these initially because I didn't see a way they could be called without being on a plannet, but I've added them now.

Mushroom (approved):

Baldie-dev (approved):

PR #190, opened 2021-04-18
Fix for drone deadlock

Baldie-dev:
This fix is for better synchronization of drones and it solves deadlocks that can occur due to the delay between host and client.

Two main situations were causing this:

1. Client sent a drone to build something and the drone will arrive and call "Build". However, we are not building it immediately on the client, we just sent a request to the host and wait for the reply, and then the building is built. So once the drone arrives, the "Build" is called, the request is sent and the drone is going back to the player. This causes that another drone was dispatched to the same building because, from the client's point of view, the building was not finished yet.
    - **Solution: transpiler to not dispatch a drone to the `buildPreview` for which client sent request to be build (another drone already reached it)**

2. Issue is with recycling the `prebuild` IDs. When you place `buildPreview` it has `Id = -1`, when that building is constructed and you place another `buildPreview`, it has again `Id = -1`.  So due to the delay, client sent a drone to the first build preview (`id: -1`) because he does not see it build yet. The host will receive info about it, but from the host point of the `id: -1` is for the second `buildPreview`, which the client does not see yet. In short: The client sent a drone to building A, but the host thinks that the drone is going to building B. And both buildings had the same ID.
    - **Solution: I have added the position as a target for the drone to check if he is going to the same `buildPreview `and we think.**


NebulaWorld/Player/DroneManager.cs
```diff
@@ -9,16 +9,26 @@ public static class DroneManager
         public static int[] DronePriorities = new int[255];
         public static Random rnd = new Random();
         public static Dictionary<ushort, List<int>> PlayerDroneBuildingPlans = new Dictionary<ushort, List<int>>();
+        public static List<int> PendingBuildRequests = new List<int>();
```
Therzok, line 12: Shouldn't this be locked?

NebulaWorld/Player/DroneManager.cs
```diff
@@ -46,5 +56,20 @@ public static int[] GetPlayerDronePlans(ushort playerId)
...
+        }
+
+        public static bool IsPendingBuildRequest(int entityId)
+        {
+            return PendingBuildRequests.Contains(entityId);
+        }
+
+        public static void RemoveBuildRequest(int entityId)
+        {
+            PendingBuildRequests.Remove(entityId);
```
Therzok, line 72: Allow notifying caller if the entry was removed.

```suggestion
        public static bool RemoveBuildRequest(int entityId)
        {
            return PendingBuildRequests.Remove(entityId);
```

NebulaWorld/Player/DroneManager.cs
Baldie-dev: I believe it is modified only from the main thread, from the `GameTick()` or from the `Update()` when packets are processed. It is purely used only on the client-side.

NebulaWorld/Player/DroneManager.cs
Baldie-dev: I am not sure if it would be helpful, maybe just to print a warning to logs that some exceptional case occured?
Since:

- Item is added to the array when the build request is sent to the host.
- item is removed from the array when the response from the host is received.

Baldie-dev: added 👍

NebulaPatcher/Patches/Transpilers/MechaDroneLogic_Patch.cs
```diff
@@ -32,6 +32,32 @@ static IEnumerable<CodeInstruction> UpdateTargets_Transpiler(IEnumerable<CodeIns
...
+            {
+                if (codes[i].opcode == OpCodes.Brtrue &&
+                    codes[i - 1].opcode == OpCodes.Callvirt &&
+                    codes[i + 1].opcode == OpCodes.Ldloc_S &&
+                    codes[i - 2].opcode == OpCodes.Ldloc_S &&
+                    codes[i + 2].opcode == OpCodes.Stloc_2)
+                {
+                    codes.InsertRange(i + 1, new CodeInstruction[] {
+                        new CodeInstruction(OpCodes.Ldloc_S, 6),
+                        new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(DroneManager), "IsPendingBuildRequest", new System.Type[] { typeof(int) })),
```
Therzok, line 54: can nameof be used here?

Therzok: Type safety.

```suggestion
                        new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(DroneManager), nameof(DroneManager.IsPendingBuildRequest), new System.Type[] { typeof(int) })),
```

NebulaWorld/Player/DroneManager.cs
Therzok: I might be thinking about it too much, but there's not much clarity or convention on when something can be accessed from Host/Client/Both and what is the source of truth for each one.

Baldie-dev: For this case of `PendingBuildRequests` it is used by clients to remember, for which prebuilds the request was sent. The issue was following: drone reached the prebuild, client sent request to build it, however from the drone point of view the prebuild still exist and still need to be built. So I have created this list to ignore prebuilds for which the client already sent request and it is just waiting for the response from server. 

But it is true, that for the "Manager" classes, it is sometimes hard to see, what is used by clients and what by the host.

NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
```diff
@@ -31,6 +33,13 @@ public void ProcessPacket(BuildEntityRequest packet, NebulaConnection conn)
                         planet.audio.Init();
                     }
 
+                //Remove building from drone queue
+                GameMain.mainPlayer.mecha.droneLogic.serving.Remove(-packet.PrebuildId);
+                planet.factory.BuildFinally(GameMain.mainPlayer, packet.PrebuildId);
+                if (!DroneManager.RemoveBuildRequest(-packet.PrebuildId))
+                {
+                    Log.Warn($"Build Request was not succesful removed.");
```
Mushroom, line 41: Typo, should be "successfully"

NebulaModel/Packets/Players/NewDroneOrderPacket.cs
```diff
@@ -8,16 +11,18 @@ public class NewDroneOrderPacket
...
-        public NewDroneOrderPacket(int planetId, int droneId, int entityId, ushort playerId, int stage, int priority)
+        public NewDroneOrderPacket(int planetId, int droneId, int entityId, ushort playerId, int stage, int priority, Vector3 entityPos)
         {
             PlanetId = planetId;
             DroneId = droneId;
             EntityId = entityId;
             PlayerId = playerId;
             Stage = stage;
             Priority = priority;
+            EntityPos = new Float3(entityPos);
```
Mushroom, line 25: Should we be consistent and use `.ToFloat3()` instead?

Mushroom (changes requested):
Only a few comments, rest looks good

NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
Baldie-dev: fixed :)

NebulaModel/Packets/Players/NewDroneOrderPacket.cs
Baldie-dev: fixed :)

hubastard (approved):

Mushroom (approved):

Issue #174, opened 2021-04-17
Soil is added to all players


Baldie-dev:
When some build building, the soil is added to all players.

I am working on the fix.

PR #191, opened 2021-04-18
Adding soil to the correct player

Baldie-dev:
This is a fix for the #174 where the soil is added to all players when the building is placed.

When the host sends the `EntityBuilt` packet to the clients, it also includes the `id` of the author who builds it, and this author will receive the soil. I add a check, that prevents adding soil if the client is not the author of the build.

output

### DIFF
--- a/NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
+++ b/NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
@@ -19,6 +19,7 @@ namespace NebulaClient.PacketProcessors.Factory.Entity
             {
                 FactoryManager.EventFromServer = true;
                 FactoryManager.EventFactory = planet.factory;
+                FactoryManager.PacketAuthor = packet.AuthorId;
 
                 // Physics could be null, if the host is not on the requested planet
                 if (packet.PlanetId != GameMain.localPlanet?.id)
@@ -42,6 +43,7 @@ namespace NebulaClient.PacketProcessors.Factory.Entity
                     planet.audio = null;
                 }
 
+                FactoryManager.PacketAuthor = -1;
                 FactoryManager.EventFromServer = false;
                 FactoryManager.EventFactory = null;
             }
--- a/NebulaPatcher/Patches/Dynamic/Player_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/Player_Patch.cs
@@ -15,7 +15,7 @@ namespace NebulaPatcher.Patches.Dynamic
         [HarmonyPatch("SetSandCount")]
         public static bool RemoveLayer_Prefix()
         {
-            //Soil should be given in singleplayer or to he player who is author of the "Build" request, or to the host if there is no author.
+            //Soil should be given in singleplayer or to the player who is author of the "Build" request, or to the host if there is no author.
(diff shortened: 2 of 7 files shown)


Baldie-dev:
+ added quick fix to load saved sand count of the player.

NebulaPatcher/Patches/Dynamic/Player_Patch.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace NebulaPatcher.Patches.Dynamic
+{
+    [HarmonyPatch(typeof(Player))]
+    class Player_Patch
+    {
+        [HarmonyPrefix]
+        [HarmonyPatch("SetSandCount")]
+        public static bool RemoveLayer_Prefix()
+        {
+            //Soil should be given in singleplayer or to he player who is author of the "Build" request, or to the host if there is no author.
```
Mushroom, line 18: Typo -> "he" should be "the"

Mushroom (approved):
Looks good other than one typo

NebulaPatcher/Patches/Dynamic/Player_Patch.cs
Baldie-dev: fixed

NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
```diff
@@ -20,6 +20,7 @@ public void ProcessPacket(BuildEntityRequest packet, NebulaConnection conn)
                 using (FactoryManager.EventFromServer.On())
                 {
                     FactoryManager.EventFactory = planet.factory;
+                    FactoryManager.PacketAuthor = packet.AuthorId;
```
Therzok, line 23: Should these always be set? Maybe could find a way to save this context.

Baldie-dev: This `AuthorId `is used to solve collisions such as: if two players build two buildings, the host will build the first one and denies the second one, and to the notification packet about the building being built, the host will attach the `AuthorId` so the correct player will remove the item from the inventory and add sand count.

The same is going for the building removal request when two players remove the same building at the same time, the host will include `authorId` for the removal request that was successful so the correct player will add item to the inventory.

Since adding and removing items from inventory + adding sand count is done in different files I have needed to add it here.

Mushroom (approved):

NebulaClient/PacketProcessors/Factory/Entity/BuildEntityRequestProcessor.cs
Therzok: Oh, I was thinking about having this exposed in the EventFromServer toggle, by creating an Author id scope or something.

Baldie-dev: If you mean something like:
```cs
using (FactoryManager.EventFromServer.On(packet.AuthorId)) {

//usage:
FactoryManager.EventFromServer.AuthorId
```
I am not exactly sure how to modify the toggle to store extra information.

Therzok: I can modify it, it would mean that toggle could store a `Stack<TMetadata>` which would be pushed to in `On` and popped when disposed. Don't worry about it too much.

PR #193, opened 2021-04-18
Change Drone conflict resolution to prioritize by distance to target

Shad0wlife:
In the unlikely case that the squared distance is exactly equal, the random priority is still utilized.

### DIFF
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -196,19 +196,37 @@ namespace NebulaWorld
                                     break;
                                 }
                             }
-                            //compare, who's drone has higher priority
-                            if (packet.Priority > priority)
+
+                            ref MechaDrone myDrone = ref myMecha.drones[droneId];
+                            //for size comparison sqrMagnitude is fine, since sqrMagnitude, magnitude and the actual distance along the curve are all strictly monotonically increasing
+                            float diff = (myDrone.position - myDrone.target).sqrMagnitude - (drone.position - drone.target).sqrMagnitude;
+                            if (diff > 0)
                             {
-                                //their drone won, my drone has to return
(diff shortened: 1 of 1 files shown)


NebulaWorld/SimulatedWorld.cs
```diff
@@ -196,20 +196,38 @@ public static void UpdateRemotePlayerDrone(NewDroneOrderPacket packet)
                                     break;
                                 }
                             }
-                            //compare, who's drone has higher priority
-                            if (packet.Priority > priority)
+
+                            ref MechaDrone myDrone = ref myMecha.drones[droneId];
+                            //for size comparison sqrMagnitude is fine, since sqrMagnitude, magnitude and the actual distance along the curve are all strictly monotonically increasing
+                            float diff = (myDrone.position - myDrone.target).sqrMagnitude - (drone.position - drone.target).sqrMagnitude;
```
Baldie-dev, line 202: there is just one situation that can cause a deadlock. If the client will be moving with high ping the host will see the client closer to the building, but from the client's point of view, the host will be closer. So nobody will sent drone to build :(.

Shad0wlife: Ah. I thought this code only ran on the host. Can I have the host only calculate the distance and then give the priority?

Baldie-dev: It is run on the client and host at the same it. I did not want to make a big overhead for the host to calculate every single drone from other players. The distance is really nice and I like it. Maybe we can keep it and just add some cooldown for the building lock.
For example: if some player will say that his drone is going there, after the 30s the lock will be lifted and any player can try it to send the drone again. 
Let me test your PR :)

Baldie-dev (approved):

Baldie-dev:
Tested and it seems to be working fine, thanks

Mushroom (approved):

PR #196, opened 2021-04-19
Ip adress sanitization for MP Menu screen

pGlase:
After being annoyed by Copy&Paste errors when copying from Discord i figured I could add this as an entry exercise.
Maybe this is a good idea to prevent A) Crashes on invalid IPs and B) Give at least a little feedback about IPv4 only.

Please let me know if this is useable for the project, I am usually a C++ dev.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
@@ -172,7 +172,7 @@ namespace NebulaPatcher.Patches.Dynamic
             int port;
 
             //remove copy and paste mistakes and update the textbox to prevent user confusion in case of invalid ip address
-            ip = SanitizeIpAddressString(ip);
+            ip = ip.Trim('\r', '\n', '\t', ' ');
             hostIPAdressInput.text = parts.Length == 1 ? ip : ip + ":" + parts[1];
 
             if (!System.Net.IPAddress.TryParse(ip, out _))
@@ -193,7 +193,6 @@ namespace NebulaPatcher.Patches.Dynamic
                 return;
             }
 
-
             // TODO: Should display a loader during the connection and only open the game once the player is connected to the server.
             multiplayerMenu.gameObject.SetActive(false);
 
@@ -207,19 +206,6 @@ namespace NebulaPatcher.Patches.Dynamic
         {
             multiplayerMenu.gameObject.SetActive(false);
             UIRoot.instance.OpenMainMenuUI();
-        }
-
-        private static string SanitizeIpAddressString(string ipAddr)
-        {
-            //remove simple copy&paste errors
-            string cleanedIpAddr = new System.Text.StringBuilder(ipAddr)
-                .Replace("\r", string.Empty)
-                .Replace("\n", string.Empty)
-                .Replace(" ", string.Empty)
-                .Replace("\t", string.Empty)
-                .ToString();
-
-            return cleanedIpAddr;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
```diff
@@ -171,6 +171,18 @@ private static void OnJoinGameButtonClick()
             string ip = parts[0];
             int port;
 
+            //remove copy and paste mistakes and update the textbox to prevent user confusion in case of invalid ip address
+            ip = SanitizeIpAddressString(ip);
```
Mushroom, line 175: I think using `String.Trim()` is probably better than a new function here

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
```diff
@@ -171,6 +171,18 @@ private static void OnJoinGameButtonClick()
...
+            hostIPAdressInput.text = parts.Length == 1 ? ip : ip + ":" + parts[1];
+
+            if (!System.Net.IPAddress.TryParse(ip, out _))
+            {
+                //log, notify user and early return to prevent crash on invalid strings
+                InGamePopup.ShowWarning("Invalid connection string", "Note: This mod currently only accepts IPv4 Addresses.", "OK");
```
Mushroom, line 181: Make sure the text in this message box is not clipped on low resolutions / high scaling, it's quite long

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
```diff
@@ -181,6 +193,7 @@ private static void OnJoinGameButtonClick()
                 return;
             }
 
+
```
Mushroom, line 196: Random newline?

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
```diff
@@ -195,5 +208,18 @@ private static void OnJoinGameBackButtonClick()
             multiplayerMenu.gameObject.SetActive(false);
             UIRoot.instance.OpenMainMenuUI();
         }
+
+        private static string SanitizeIpAddressString(string ipAddr)
```
Mushroom, line 212: See comment above about using `String.Trim()` instead

Mushroom (changes requested):
Ahh, as an embedded dev I understand your approach completely, but this is not the C# way :)

I have suggested a better alternative

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
pGlase: I just tried this on 640x480 on my 1440p display, i don't recommend it, but at least it isn't clipping ;)
grafik

Mushroom: Alright then, that's one concern alleviated!

Mushroom (approved):

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
Therzok: Trim is identical to Trim(whitespace)

NebulaPatcher/Patches/Dynamic/UIMainMenu_Patch.cs
```diff
@@ -171,6 +171,18 @@ private static void OnJoinGameButtonClick()
             string ip = parts[0];
             int port;
 
+            //remove copy and paste mistakes and update the textbox to prevent user confusion in case of invalid ip address
+            ip = ip.Trim('\r', '\n', '\t', ' ');
```
Therzok, line 175: These are all already considered whitespace by c#

PR #212, opened 2021-04-22
Bug Fix: Prebuild indexes are now taken from the game logic

Baldie-dev:
I found another issue when drone research is unlocked, client for some reason starts sending "BuildFinally()" packet twice, even after reconnecting.

After 10 minutes of testing, no issue observed

Following changes were applied:

- Prebuild ID is not calculated anymore, but it is taken from the game logic after prebuild is actually created.
- Some static fields were not reseted in the `DroneManager` so I have added `Initialize()` method to reset all static fields.
- Added check, so the second "BuildFinally()" call will not do anything, it will just produce a warning on the host. (I will check in the future why this is happening).

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/PlanetFactory_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/PlanetFactory_Patch.cs
@@ -13,21 +13,19 @@ namespace NebulaPatcher.Patches.Dynamic
     [HarmonyPatch(typeof(PlanetFactory))]
     class BuildFinally_patch
     {
-        [HarmonyPrefix]
-        [HarmonyPatch("AddPrebuildDataWithComponents")]
-        public static bool AddPrebuildDataWithComponents_Prefix(PlanetFactory __instance, PrebuildData prebuild)
+        [HarmonyPostfix]
+        [HarmonyPatch("AddPrebuildData")]
+        public static void AddPrebuildData_Prefix(PlanetFactory __instance, PrebuildData prebuild, ref int __result)
         {
             if (!SimulatedWorld.Initialized)
-                return true;
+                return;
 
             // If the host game called the method, we need to compute the PrebuildId ourself
             if (LocalPlayer.IsMasterClient && !FactoryManager.EventFromClient)
             {
-                int nextPrebuildId = FactoryManager.GetNextPrebuildId(__instance);
-                FactoryManager.SetPrebuildRequest(__instance.planetId, nextPrebuildId, LocalPlayer.PlayerId);
+               
+                FactoryManager.SetPrebuildRequest(__instance.planetId, __result, LocalPlayer.PlayerId);
             }
-
-            return true;
         }
 
         [HarmonyPrefix]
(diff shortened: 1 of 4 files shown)


NebulaWorld/Player/DroneManager.cs
```diff
@@ -13,6 +13,14 @@ public static class DroneManager
         public static List<int> PendingBuildRequests = new List<int>();
         public static Dictionary<ushort, Vector3> CachedPositions = new Dictionary<ushort, Vector3>();
 
+        public static void Initialize()
+        {
+            DronePriorities = new int[255];
```
Therzok, line 18: They're already initialized in the constructor?

Therzok: Why is an explicit initialize call needed here?

NebulaPatcher/Patches/Dynamic/PlanetFactory_Patch.cs
```diff
@@ -13,21 +13,19 @@ namespace NebulaPatcher.Patches.Dynamic
     [HarmonyPatch(typeof(PlanetFactory))]
     class BuildFinally_patch
     {
-        [HarmonyPrefix]
-        [HarmonyPatch("AddPrebuildDataWithComponents")]
-        public static bool AddPrebuildDataWithComponents_Prefix(PlanetFactory __instance, PrebuildData prebuild)
+        [HarmonyPostfix]
+        [HarmonyPatch("AddPrebuildData")]
+        public static void AddPrebuildData_Prefix(PlanetFactory __instance, PrebuildData prebuild, ref int __result)
```
hubastard, line 18: Should rename the method to `AddPrebuildData_Postfix`

NebulaWorld/Player/DroneManager.cs
hubastard: It's probably to reset all the collections each time a player host or join a new game without restarting the game.

hubastard (approved):

NebulaWorld/Player/DroneManager.cs
Therzok: I would say that removing the initializers in the constructor (the ones above) is the best way to figure out if something uses the service before initialization. Or have everything done in a session object that can be referred to for all the current session managers.

NebulaPatcher/Patches/Dynamic/PlanetFactory_Patch.cs
Baldie-dev: fixed :)

NebulaWorld/Player/DroneManager.cs
Baldie-dev: Since the fields above are static, they are persistent when user disconnect and connect again. This servers as "Cleaning", so every time user connects to the game, the arrays will be fresh and clean.

Therzok: I put it on my list to look at making a `Session` object that's refreshed on every connection.

PR #214, opened 2021-04-23
Ping Indicator for clients

Baldie-dev:
I have added ping indicator for clients. 

Since it is my first UI thingy, all comments welcomed:
image

### DIFF
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -83,7 +83,7 @@ namespace NebulaClient
         {
             if (pingIndicator == null)
             {
-                GameObject targetObject = GameObject.Find("fps-ups-text");
+                GameObject targetObject = GameObject.Find("label");
                 pingIndicator = GameObject.Instantiate(targetObject, targetObject.transform.parent.parent).GetComponent<Text>();
                 pingIndicator.gameObject.name = "Ping Indicator";
                 pingIndicator.alignment = TextAnchor.UpperLeft;
@@ -92,6 +92,7 @@ namespace NebulaClient
                 rect.anchorMin = new Vector2(0f, 1f);
                 rect.offsetMin = new Vector2(10f, -100f);
                 pingIndicator.text = "";
+                pingIndicator.fontSize = 14;
             }
         }
 
--- a/NebulaClient/MultiplayerClientSession.cs
+++ b/NebulaClient/MultiplayerClientSession.cs
@@ -27,6 +27,7 @@ namespace NebulaClient
         private float pingTimer = 0f;
         private float pingTimestamp = 0f;
         private Text pingIndicator;
+        private int previousDelay = 0;
 
         public NetPacketProcessor PacketProcessor { get; protected set; }
         public bool IsConnected { get; protected set; }
@@ -81,7 +82,8 @@ namespace NebulaClient
 
         public void DisplayPingIndicator()
         {
-            if (pingIndicator == null)
+            GameObject previousObject = GameObject.Find("Ping Indicator");
+            if (previousObject == null)
             {
                 GameObject targetObject = GameObject.Find("label");
                 pingIndicator = GameObject.Instantiate(targetObject, targetObject.transform.parent.parent).GetComponent<Text>();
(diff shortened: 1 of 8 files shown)


Baldie-dev:
Font changed to:
image

hubastard (approved):

NebulaClient/MultiplayerClientSession.cs
```diff
@@ -73,6 +79,23 @@ private void ConnectInternal()
             LocalPlayer.SetNetworkProvider(this);
         }
 
+        public void DisplayPingIndicator()
+        {
+            if (pingIndicator == null)
+            {
+                GameObject targetObject = GameObject.Find("label");
+                pingIndicator = GameObject.Instantiate(targetObject, targetObject.transform.parent.parent).GetComponent<Text>();
```
Therzok, line 87: Can this be a more specific UI gameobject? I think UIRoot or UIGame has members for important layers.

NebulaClient/MultiplayerClientSession.cs
```diff
@@ -82,6 +105,7 @@ void Disconnect()
         public void DestroySession()
         {
             Disconnect();
+            Destroy(pingIndicator.gameObject);
```
Therzok, line 108: Or just enabled=false for reuse?

NebulaClient/MultiplayerClientSession.cs
```diff
@@ -111,6 +135,12 @@ public void Reconnect()
             ConnectInternal();
         }
 
+        public void UpdatePingIndicator()
+        {
+            float delay = (Time.time - pingTimestamp);
+            pingIndicator.text = $"Ping: {(int)(delay*1000)}ms";
```
Therzok, line 141: Only update if it changed?

NebulaClient/MultiplayerClientSession.cs
Baldie-dev: `transform.parent.parent` is `UIGame` object.
 But I can also change it to:
 ```cs
 GameObject.Instantiate(targetObject, GameObject.Find("in game").transform)
 ```
 But it would be same.

NebulaClient/MultiplayerClientSession.cs
Baldie-dev: yup, I can do that :)

NebulaClient/MultiplayerClientSession.cs
Baldie-dev: This should be updated every second, so I do not think there is much impact. But I will add that condition.

Baldie-dev: done :)

NebulaClient/MultiplayerClientSession.cs
Baldie-dev: done :)

NebulaClient/MultiplayerClientSession.cs
Therzok: Or you can use `UIRoot.instance.uiGame` then?

Therzok: The reasoning is that if it is removed from the game, you get a compile error.

You don't get it if the unity scene is reorganized

Baldie-dev: true,

Baldie-dev: fixed :)

Therzok (approved):

hubastard:
Did you test what happen if the fps meter is displayed or if you have a research started. I'm wondering if it will overlap with other UI stuff. 

Maybe moving the label to be under the player name on the top right would be better?

Mushroom:
I agree, this should be tested with the FPS meter too - the under-username location seems better (or maybe next to version?)

Baldie-dev:
I can move it more down, I have never see the FPS meter :D Is there any shortcut to display it?

hubastard:
It shows itself when the fps drops under 30fps. You may be better of making a patch that keep it shown 😂

Baldie-dev:
Offset added:
image

hubastard (approved):

PR #215, opened 2021-04-24
Disable Nagle algorithm to allow for faster packet sending

Therzok:
Nagle tries to batch multiple packets together, which can potentially delay the messages from being sent. So over here, just send them as they're needed


Therzok:
Don't merge this yet. I'm experimenting with buffer sizes for packets.

hubastard (approved):

Therzok:
Sooo, there's also a way to send ping messages via websocketsharp and there's a way to compress packets. Let me try and figure those out first.

Therzok:
Can be merged.

Issue #243, opened 2021-04-28
Client are sometimes teleported to the wrong planet when rejoining a previous host.


hubastard:
Sometime when a client joins a server for the second time, the client is teleported to the wrong planet.

hubastard:
After some digging it looks like the planetID are not the same on each players :S

hubastard:
This will be fixed in the PR #244. The issue was caused by the fact that we were loading the wrong server save file. So, if the host had 2 different games on disk and the same client played in both of them, sometimes, it was loading the wrong server save and was sending the client state from the other save file. Since, planetID are similar from one game to the other, it means that the client was teleported to the planet that had the same planetID in their other game.

PR #244, opened 2021-04-28
Fix save game loading issues

### LINKED ISSUE #242: Clients that join back the server lose their state.
In some conditions clients that join back a server loses their inventory.

1. Host a game
2. Join with a client
3. Move both players out of the pod
4. Save the game on the host
5. Make the client Quit the game
6. (important) Make the host Quit the game
7. Host back the game that you saved
8. Join back with the client
9. Notice that the host position is correct but the client now a fresh new player spawned back on the pod

hubastard:
This PR fixes #242 and #243, which are the issues that we had where reconnecting players was sometimes joining as new players or on another planet.

We had 2 main issues:
1. We were not loading the server save file when we were hosting a game from the Load Game Window which was causing all the players to join with no player data. (This was not occurring if the host stayed in game while the clients reconnected)
2. The other issue with the players that were spawning on another planet was caused by the fact that we were not handling the save game loading of AutoSaves properly. So we were always reloading the last autosave instead of the actually loading the game that the host had chosen in the Load Game Window.

This PR also fixes the AutoSave rotation that was not handled previously and I also removed the Host button from the Esc menu.

### DIFF
--- a/NebulaHost/SaveManager.cs
+++ b/NebulaHost/SaveManager.cs
@@ -1,4 +1,5 @@
-﻿using NebulaModel.DataStructures;
+﻿using HarmonyLib;
+using NebulaModel.DataStructures;
 using NebulaModel.Networking.Serialization;
 using NebulaModel.Utils;
 using NebulaWorld;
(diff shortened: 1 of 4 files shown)


hubastard:
I will close it right now to make sure that this fix is available as soon as possible. But if you comments you can add them and I will fix them in a subsequent PR.

PR #254, opened 2021-04-30
Added a Multiplayer Options tab in Settings UI

hubastard:
- Added a new `MultiplayerOptions` class that let us add some settings that we want to expose to the in game settings menu. 
- Properties are automatically added to the settings UI and linked with `onChange` events.
- For now automatic UI generation supports the following data types `int`, `float`, `bool`, `string` and `enum`
- All properties that you want to expose in the UI most have a `UIControl` attribute
- If you want to have a Slider instead of an input field for a number, you can add the `UIRange` attribute on your property.

[see video](https://youtu.be/BJR105BKVfE)


NebulaModel/Attributes/UIControlAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace NebulaModel.Attributes
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class UIControlAttribute : Attribute
+    {
+        public readonly string displayName;
+
+        public UIControlAttribute(string displayName)
+        {
+            this.displayName = displayName;
```
Therzok, line 12: Might be nitpicky, but System.ComponentModel has these attributes.

TheRealMarine (commented):
I don't see a max port for the Options. Every port over 65.535 does not work as far as i know. Some people might enter larger numbers and report it as a bug.

NebulaModel/Attributes/UIRangeAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace NebulaModel.Attributes
+{
+    [AttributeUsage(AttributeTargets.Property, AllowMultiple = false)]
+    public class UIRangeAttribute : Attribute
+    {
+        public readonly float min;
+        public readonly float max;
+
+        public UIRangeAttribute(float min, float max)
```
Therzok, line 11: Can use: https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.rangeattribute?view=netframework-3.5

NebulaModel/Attributes/UIControlAttribute.cs
Therzok: Could use https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.displaynameattribute?view=netframework-3.5 instead.

NebulaModel/Config.cs
```diff
@@ -1,11 +1,66 @@
...
+
+        public static bool SaveOptions()
+        {
+            if (Options == null)
+            {
+                Log.Error($"Cannot save {OPTION_SAVE_FILE}, is null");
+                return false;
+            }
+
+            string path = GameConfig.gameDocumentFolder + OPTION_SAVE_FILE;
```
Therzok, line 49:
```suggestion
            string path = Path.Combine(GameConfig.gameDocumentFolder, OPTION_SAVE_FILE);
```

NebulaModel/Config.cs
```diff
@@ -1,11 +1,66 @@
...
         public static PluginInfo ModInfo { get; set; }
         public static string ModVersion => ThisAssembly.AssemblyInformationalVersion;
-        public static int DefaultPort => 8469;
+        public static MultiplayerOptions Options { get; set; }
+
+        public static bool LoadOptions()
+        {
+            Options = new MultiplayerOptions();
+
+            string path = GameConfig.gameDocumentFolder + OPTION_SAVE_FILE;
```
Therzok, line 20:
```suggestion
            string path = Path.Combine(GameConfig.gameDocumentFolder, OPTION_SAVE_FILE);
```

NebulaModel/Attributes/AttributeExtension.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public static class AttributeExtension
+    {
+        public static T GetCustomAttribute<T>(this PropertyInfo property) where T : Attribute
+        {
+            return property.GetCustomAttributes(true)
+                .Select(att => att as T)
+                .Where(att => att != null)
+                .FirstOrDefault();
+        }
```
Therzok, line 15: Use MemberInfo for completeness and 
```suggestion
        public static T GetCustomAttribute<T>(this MemberInfo member) where T : Attribute
        {
            return member.GetCustomAttributes(true)
                .OfType<T>()
                .FirstOrDefault();
        }
```

NebulaPatcher/Patches/Dynamic/UIOptionWindow_Patch.cs
```diff
@@ -0,0 +1,279 @@
...
+
+            Tweener[] newContents = CollectionExtensions.AddToArray(tabTweeners, multiplayerContent.GetComponent<Tweener>());
+            AccessTools.Field(__instance.GetType(), "tabTweeners").SetValue(__instance, newContents);
+            UIButton[] revertButtons = AccessTools.Field(__instance.GetType(), "revertButtons").GetValue(__instance) as UIButton[];
+            RectTransform revertButton = multiplayerContent.Find("revert-button").GetComponent<RectTransform>();
+            UIButton[] newRevertButtons = CollectionExtensions.AddToArray(revertButtons, revertButton.GetComponent<UIButton>());
+            AccessTools.Field(__instance.GetType(), "revertButtons").SetValue(__instance, newRevertButtons);
+
+            // Find control templates
+            foreach (RectTransform child in multiplayerContent)
```
Therzok, line 71: Would probably be a bit simpler if you have a Panel holding onto all the multiplayerContent children that need removing.

Therzok: Then it becomes `Object.Destroy(panel)`

NebulaPatcher/Patches/Dynamic/UIOptionWindow_Patch.cs
```diff
@@ -0,0 +1,279 @@
...
+        [HarmonyPatch("TempOptionToUI")]
+        public static void TempOptionToUI_Postfix()
+        {
+            List<PropertyInfo> properties = AccessTools.GetDeclaredProperties(typeof(NebulaModel.MultiplayerOptions));
+            foreach (var prop in properties)
+            {
+                if (tempToUICallbacks.ContainsKey(prop.Name))
+                {
+                    tempToUICallbacks[prop.Name]();
+                }
```
Therzok, line 129: A bit easier to follow:

```suggestion
                if (tempToUICallbacks.TryGetValue(prop.Name, out var callback))
                {
                    callback();
                }
```

NebulaPatcher/Patches/Dynamic/UIOptionWindow_Patch.cs
```diff
@@ -0,0 +1,279 @@
...
+
+        private static void CreateBooleanControl(UIControlAttribute control, PropertyInfo prop, Vector2 anchorPosition)
+        {
+            RectTransform element = Object.Instantiate(checkboxTemplate, multiplayerContent, false);
+            SetupUIElement(element, control, prop, anchorPosition);
+            UIToggle toggle = element.GetComponentInChildren<UIToggle>();
+            toggle.toggle.onValueChanged.RemoveAllListeners();
+            toggle.toggle.onValueChanged.AddListener((value) => { prop.SetValue(tempMultiplayerOptions, value, null); });
+
+            tempToUICallbacks[prop.Name] = () =>
```
Therzok, line 178: This might end up with a case where, due to GameObject.Destroy(), this dictionary will contain a null value. This dictionary should be cleared when the multiplayer content widgets are removed.

NebulaModel/MultiplayerOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using NebulaModel.Attributes;
+using System;
+
+namespace NebulaModel
+{
+    [System.Serializable]
+    public class MultiplayerOptions : ICloneable
+    {
+        [UIControl("Host Port")]
+        public int HostPort { get; set; } = 8469;
```
Therzok, line 10: As @TheRealMarine mentioned, this should be an ushort, with `Range(1, ushort.MaxValue)`

NebulaPatcher/Patches/Dynamic/UIOptionWindow_Patch.cs
hubastard: This ain't my UI really. The code in OnCreate is  essentially cloning existing game UI to make sure that I have the same look, so the destroy here is to flush what I don't want to keep from the cloned element.

Therzok: Yeah, what I'm suggesting, create a `Panel container` gameobject with just a rect transform and shove all the child gameobjects there.

Then `Destroy(container)` will do everything for you, even if some _other_ code attaches gameobjects in it.

NebulaModel/Attributes/UIRangeAttribute.cs
hubastard: Oh wow I didn't knew those existed. I'll change the code to use them

hubastard: humm, I'll keep my UIRangeAttribute, otherwise we need to include the entire DataAnnotation dll with the mod. Too overkill for our usage.

NebulaPatcher/Patches/Dynamic/UIOptionWindow_Patch.cs
hubastard: Humm, not sure I see why it would be better. Because instead of doing:
- clone existing tab container
- destroy all gameObjects in it.

It would be:
- clone existing tab container
- create custom container
- iterate over all children of the tab container to set our own container as parent
- destroy the container

Therzok: Probably force of habit here, in case any other content needed to be cloned, you would generally need to care only about that one and Destroy it. But `YDNIN` applies here. Not needed.

Therzok (approved):

PR #258, opened 2021-04-30
When a client loads in on a different planet than host, it does not …

Jon-Myrstad:
have real PlanetData for other planets. PlanetData.vegePool is null and was causing a nullref exception. This checks if vegePool is null before attempting RemoveVegeWithComponents.

I'm not entirely sure if a client should have received this or not at this point as I don't know the intent of what is planned to be host-authorative and re-loaded on demand and what is supposed to be synced once and then simulated on clients.

Obviously the client doesn't actually need this packet, since it has no PlanetData for the planet it's trying to remove vegetation on, but I think that is a seperate issue in trying to reduce packet count.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -366,3 +366,6 @@ MigrationBackup/
 
 # Fody - auto-generated XML schema
 FodyWeavers.xsd
+
+# Jetbrains IDE working folder
+.idea/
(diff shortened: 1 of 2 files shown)


hubastard (approved):

Baldie-dev:
I think that client needs to get this packet to remove vege data from the planet.
For example:
1. client visits the planet
2. client leaves the planet
3. host removes vege data from the planet. 
4. since the client has cached the planet, it needs to remove the vege data locally too. Since when he comes back, he will not ask the server for the vege update, but it will load vege data from the local storage.

I will check why there was a null pointer exception.

But this commit should not have impact on it.

PR #259, opened 2021-04-30
Changed Port from Integer to ushort

TheRealMarine:


NebulaModel/MultiplayerOptions.cs
```diff
@@ -7,7 +7,7 @@ namespace NebulaModel
     public class MultiplayerOptions : ICloneable
     {
         [UIControl("Host Port")]
-        public int HostPort { get; set; } = 8469;
+        public ushort HostPort { get; set; } = 8469;
```
Therzok, line 10: Afaik, port 0 shouldn't be allowed either.

Therzok (approved):

PR #269, opened 2021-05-01
Refactored large part of the UIStationWindow and UIStationStorage sync logic

Jon-Myrstad:
Updating PLS's wasn't working because a lot of checks were being made on GalacticTransport.stationPool.Length even when sending data about a non-stellar station. Instead of trying to force in fixes for those specific cases I did some refactoring of the UIStationWindow and UIStationStorage functionality.

- Subscriptions to Station updates are now based on planetId, stationId and stationGid to make sure subscriptions work on both ILS and PLS stations.
- With the subscription change packets also send planetId, stationId and stationGid along with their changes.
- Refactored UIStation packets to follow .editorconfig.
- Changed processors and logic that follows processors to resolve which station were making changes to based on planetId, stationId and stationGid.
- Removed a lot of reduntant Null checks by doing more checks further up in StationUIManager's UpdateSettingsUI and UpdateSettingsUIBackground methods.

### DIFF
--- a/NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
+++ b/NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
@@ -4,6 +4,7 @@ using NebulaModel.Packets.Logistics;
 using NebulaModel.Packets.Processors;
 using NebulaWorld.Logistics;
 using HarmonyLib;
+using NebulaModel.Logger;
 
 /*
  * When the client opens the UI of a station (ILS/PLS/Collector) the contents gets updated and shown to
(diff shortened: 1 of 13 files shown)


NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
```diff
@@ -24,13 +24,24 @@ public void ProcessPacket(StationUI packet, NebulaConnection conn)
         {
             Player player = playerManager.GetPlayer(conn);
             // if a user adds/removes a ship, drone or warper or changes max power input broadcast to everyone.
-            if((packet.settingIndex == StationUI.UIsettings.MaxChargePower || packet.settingIndex == StationUI.UIsettings.setDroneCount || packet.settingIndex == StationUI.UIsettings.setShipCount || packet.settingIndex == StationUI.UIsettings.setWarperCount) && player != null && StationUIManager.UpdateCooldown == 0)
+            if (
+                (
+                    packet.SettingIndex == StationUI.EUISettings.MaxChargePower
```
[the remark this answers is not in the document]
Jon-Myrstad: Most likely my IDE doing something for me.

Moved the first null-check and "Cooldown"-check up above all the Enums.

Not sure about the trailing bracket, I think this is more readable than two brackets on the same line.

Mushroom (approved):
LGTM, but @sp00ktober needs to approve this before merge

NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
```diff
@@ -17,61 +18,59 @@ public class StationUIInitialSyncProcessor: IPacketProcessor<StationUIInitialSyn
     {
         public void ProcessPacket(StationUIInitialSync packet, NebulaConnection conn)
         {
-            StationComponent[] gStationPool = null;
-            if(packet.planetId == 0)
+            StationComponent stationComponent = null;
+            StationComponent[] gStationPool = GameMain.data.galacticTransport.stationPool;
+            StationComponent[] stationPool = GameMain.data.galaxy.PlanetById(packet.PlanetId).factory.transport.stationPool;
```
sp00ktober, line 23: Does this NRE when the client did not have the PlanetFactory loaded? (this is the case for planets he did not visit yet)

sp00ktober (changes requested):
Only one small thing :)

NebulaClient/PacketProcessors/Logistics/StationUIInitialSyncProcessor.cs
Jon-Myrstad: This is when the client gets a response to opening a UIStationWindow, no?

So he must be on the planet. I could add ? before every period, but I don't see how it could happen.

sp00ktober: Ah yes you are right, my bad then :D

There should not be a way for a client to not have the factory loaded, even if he opens the ui and then leaves the planet before receiving this packet he still has the factory as he once was on the planet.

sp00ktober (approved):

PR #270, opened 2021-05-01
When a fake ILS station is created through ILSShipManager we create them with ShipRenderers...

Jon-Myrstad:
... and UIShipRenderers, this was not being done on ILS's that are created on initial StationPoolSync causing errors and non-functional renderers for clients, when LogisticShipRenderer::Update gets called on the station.


sp00ktober (approved):
Looks good to me :D

Issue #260, opened 2021-04-30
Using the "Clear and Reset" button for a StationStorage slot on a client makes the items that...


Jon-Myrstad:
... usually would be put into the player's inventory vanish

Issue #257, opened 2021-04-30
Manually inserting/removing items from IPS/PLS does not sync between host/client


Jon-Myrstad:
Host is top left, client is bottom right

Built from master after #86 was merged.

https://www.youtube.com/watch?v=zJzfgajR_Qk

PR #271, opened 2021-05-01
fix for #257 and #260

sp00ktober:
should fix both issues.

### DIFF
--- a/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
+++ b/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
@@ -36,16 +36,8 @@ namespace NebulaHost.PacketProcessors.Logistics
                     foreach (var kvp in connectedPlayers)
                     {
                         Player p = kvp.Value;
-                        if (p.Connection == conn)
-                        {
-                            packet.ShouldMimic = true;
-                            p.SendPacket(packet);
-                            packet.ShouldMimic = false;
-                        }
-                        else
-                        {
-                            p.SendPacket(packet);
-                        }
+                        packet.ShouldMimic = p.Connection == conn;
+                        p.SendPacket(packet);
                     }
                 }
             }
@@ -66,14 +58,11 @@ namespace NebulaHost.PacketProcessors.Logistics
                 {
                     if(subscribers[i] != null)
                     {
-                        if(subscribers[i] == conn)
-                        {
-                            /*
-                             * as we block the normal method for the client he must run it once he receives this packet.
-                             * but only the one issued the request should do it, we indicate this here
-                             */
-                            packet.ShouldMimic = true;
-                        }
+                       /*
+                        * as we block the normal method for the client he must run it once he receives this packet.
+                        * but only the one issued the request should do it, we indicate this here
+                        */
+                        packet.ShouldMimic = subscribers[i] == conn;
                         subscribers[i].SendPacket(packet);
                     }
                 }
--- a/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
+++ b/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
@@ -27,8 +27,7 @@ namespace NebulaHost.PacketProcessors.Logistics
                 (packet.SettingIndex == StationUI.EUISettings.MaxChargePower
                  || packet.SettingIndex == StationUI.EUISettings.SetDroneCount
                  || packet.SettingIndex == StationUI.EUISettings.SetShipCount
-                 || packet.SettingIndex == StationUI.EUISettings.SetWarperCount
-                 )
+                 || packet.SettingIndex == StationUI.EUISettings.SetWarperCount)
                 )
             {
                 // this is the SendPacketToAllPlayers() logic but we need to set the mimic flag here.
@@ -40,8 +39,13 @@ namespace NebulaHost.PacketProcessors.Logistics
                         if (p.Connection == conn)
                         {
                             packet.ShouldMimic = true;
+                            p.SendPacket(packet);
+                            packet.ShouldMimic = false;
                         }
-                        p.SendPacket(packet);
+                        else
+                        {
+                            p.SendPacket(packet);
+                        }
                     }
                 }
             }
--- a/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
+++ b/NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
@@ -22,11 +22,31 @@ namespace NebulaHost.PacketProcessors.Logistics
         }
         public void ProcessPacket(StationUI packet, NebulaConnection conn)
         {
-            Player player = playerManager.GetPlayer(conn);
             // if a user adds/removes a ship, drone or warper or changes max power input broadcast to everyone.
-            if((packet.settingIndex == StationUI.UIsettings.MaxChargePower || packet.settingIndex == StationUI.UIsettings.setDroneCount || packet.settingIndex == StationUI.UIsettings.setShipCount || packet.settingIndex == StationUI.UIsettings.setWarperCount) && player != null && StationUIManager.UpdateCooldown == 0)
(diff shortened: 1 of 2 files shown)


NebulaWorld/Logistics/StationUIManager.cs
```diff
@@ -471,7 +472,7 @@ private static void UpdateStorageUI(StationUI packet)
 
                 using (ILSShipManager.PatchLockILS.On())
                 {
-                    pData.factory.transport.SetStationStorage(id, packet.storageIdx, packet.itemId, packet.itemCountMax, packet.localLogic, packet.remoteLogic, null);
+                    pData.factory.transport.SetStationStorage(id, packet.storageIdx, packet.itemId, packet.itemCountMax, packet.localLogic, packet.remoteLogic, (packet.shouldMimick == true) ? GameMain.mainPlayer : null);
```
hubastard, line 475: `(packet.shouldMimick == true)` can only be `packet.shouldMimick`

NebulaWorld/Logistics/StationUIManager.cs
```diff
@@ -153,7 +153,8 @@ private static void UpdateSettingsUIBackground(StationUI packet, PlanetData pDat
             // only host should update everything
             if (!LocalPlayer.IsMasterClient)
             {
-                return;
+                // this is changed in pr #269 anyways to reflect this logic (update for everyone, not just host. after all why should'nt we)
+                //return;
             }
```
hubastard, line 158: Should remove entire if, since its not used anymore

sp00ktober: Yes, wanted to do the merge conflict resolve later

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
```diff
@@ -23,27 +23,33 @@ public StationUIBroadcaster()
...
                         {
-                            packet.shouldMimick = true;
+                            packet.ShouldMimic = true;
                         }
                         p.SendPacket(packet);
                     }
                 }
             }
-            else if(packet.settingIndex == StationUI.UIsettings.addOrRemoveItemFromStorageResp)
+            else if (packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageResponse)
```
sp00ktober, line 48: Check for item changes here and not in the if() above because we only need to sync item changes to players on the same planet.

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
```diff
@@ -22,18 +22,37 @@ public StationUIBroadcaster()
...
-                 || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageRequest
-                 || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageResponse)
+                 )
                 )
             {
-                playerManager.SendPacketToAllPlayers(packet);
+                // this is the SendPacketToAllPlayers() logic but we need to set the mimic flag here.
+                using (playerManager.GetConnectedPlayers(out var connectedPlayers))
+                {
+                    foreach (var kvp in connectedPlayers)
```
Mushroom, line 37: Can use `connectedPlayers.Values` here rather than looping the keypairs

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
```diff
@@ -22,18 +22,37 @@ public StationUIBroadcaster()
...
             // if a user adds/removes a ship, drone or warper or changes max power input broadcast to everyone.
-            if (player != null && StationUIManager.UpdateCooldown == 0 &&
+            if (StationUIManager.UpdateCooldown == 0 &&
                 (packet.SettingIndex == StationUI.EUISettings.MaxChargePower
                  || packet.SettingIndex == StationUI.EUISettings.SetDroneCount
                  || packet.SettingIndex == StationUI.EUISettings.SetShipCount
                  || packet.SettingIndex == StationUI.EUISettings.SetWarperCount
-                 || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageRequest
-                 || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageResponse)
+                 )
```
Mushroom, line 31: Don't like the lone bracket here, looks better on the line above as was before

Mushroom (changes requested):

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
Therzok: That's actually inefficient:
https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1841

Therzok: You allocate a wrapper collection for the values of the dictionary on every access:
https://github.com/Unity-Technologies/mono/blob/e20cf158facacfe732f4b53c0501e7263e114efd/mcs/class/corlib/System.Collections.Generic/Dictionary.cs#L623

Also, you can mistakenly escape `connectedPlayers.Values`, which was not possible before with the local connectedPlayers variable.

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
Therzok: Code style rule on this should be enforced. The advantage of this format is that you don't add extra diff noise by adding new conditions.

Therzok: Adding them back produces
```diff
                 || packet.SettingIndex == StationUI.EUISettings.SetWarperCount
+                || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageRequest
+                || packet.SettingIndex == StationUI.EUISettings.AddOrRemoveItemFromStorageResponse
				)

File: NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs [hunk 6]
```diff
@@ -22,18 +22,37 @@ public StationUIBroadcaster()
...
-                playerManager.SendPacketToAllPlayers(packet);
+                // this is the SendPacketToAllPlayers() logic but we need to set the mimic flag here.
+                using (playerManager.GetConnectedPlayers(out var connectedPlayers))
+                {
+                    foreach (var kvp in connectedPlayers)
+                    {
+                        Player p = kvp.Value;
+                        if (p.Connection == conn)
+                        {
+                            packet.ShouldMimic = true;
```
Therzok, line 42: This value is never reset? So, if we have 3 players:
```
p1
p2 == conn
p3
```

Both p2 and p3 will have ShouldMimic = true, but doesn't look right.

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
Therzok: You might want the key anyway, logging, etc. The fact that there's no logging in the codebase right now is worrying (yes, I know, console logging is slow, allocations in unity's backend, etc, but you don't need to use Unity's logging system.

sp00ktober: So should i change this or leave it as is?

Therzok: Leave as-is imo.

NebulaHost/PacketProcessors/Logistics/StationUIBroadcaster.cs
```diff
@@ -22,18 +22,41 @@ public StationUIBroadcaster()
...
+                            p.SendPacket(packet);
+                            packet.ShouldMimic = false;
+                        }
+                        else
+                        {
+                            p.SendPacket(packet);
+                        }
```
Therzok, line 48: A bit simpler:

```suggestion
                        Player p = kvp.Value;
                        packet.ShouldMimic = p.Connection == conn;
                        p.SendPacket(packet);
```

Mushroom (approved):

PR #279, opened 2021-05-03
Bug Fix: Build preview sometimes does not disappear from the clients.

Baldie-dev:
I am not sure why, but in the last PR #272 I have introduced a bug, that sometimes caused that BuildPreview was not removed.

This small change will fix it.

### DIFF
--- a/NebulaClient/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
+++ b/NebulaClient/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
@@ -78,9 +78,6 @@ namespace NebulaClient.PacketProcessors.Factory.Entity
                 if (packet.AuthorId == LocalPlayer.PlayerId)
                 {
                     pab.AfterPrebuild();
-                } else
-                {
-                    pab.buildPreviews = tmpList;
                 }
 
                 //Revert changes back
@@ -89,6 +86,7 @@ namespace NebulaClient.PacketProcessors.Factory.Entity
                 pab.waitConfirm = tmpConfirm;
                 pab.previewPose.position = tmpPos;
                 pab.previewPose.rotation = tmpRot;
+                pab.buildPreviews = tmpList;
 
                 FactoryManager.TargetPlanet = FactoryManager.PLANET_NONE;
             }


Mushroom:
What is the effect of moving this list outside of the scope of the if statement? Why was it moved in in the first place?

Baldie-dev:
My initial thought was to help the client to build continuously by not replacing the `buildPreviews` if the client is author of the build: `pab.buildPreviews = tmpList;`

But that has a deeper impact due to the other variables that can be set. It is working fine with belts, but it does not work fine with normal buildings. So this is just reverting that idea back.

Mushroom (approved):
Makes sense. Perhaps this could be selectively applied to belts in the future?

NebulaClient/PacketProcessors/Factory/Entity/CreatePrebuildsRequestProcessor.cs
```diff
@@ -89,6 +86,7 @@ public void ProcessPacket(CreatePrebuildsRequest packet, NebulaConnection conn)
                 pab.waitConfirm = tmpConfirm;
                 pab.previewPose.position = tmpPos;
                 pab.previewPose.rotation = tmpRot;
+                pab.buildPreviews = tmpList;
```
Therzok, line 89: Isn't this because there are multiple players writing to the shared buildPreviews? all this code should be done under some kind of locking.

Baldie-dev: No no, this is for leveraging the original code for creating the Prebuilds. When the client receives a packet from host to create prebuilds from previews, in short, it will do this:
```cs
tmpList = pab.buildPreviews;    //store current previews
pab.buildPreviews = packet.buildPreviews;  //set previews from packet
pab.CreatePrebuilds();     //native code to create prebuilds from previews
pab.buildPreviews = tmpList;    //restore current previews
```

Baldie-dev: The idea in the previous PR that introduces the bug was to not restoring `buildpreviews` if the client is the author of the building. But it does not work correctly.

hubastard (approved):

Issue #280, opened 2021-05-03
When host load game from file, the prebuilds cannot be built.


Baldie-dev:
The steps to reproduce:
1. host game, create bug of prebuilds
2. Save the game before the prebuilds are build
3. Host and load the game
4. Prebuilds will not be build anymore due to the check for prebuild requests:

```cs
 if (!FactoryManager.ContainsPrebuildRequest(__instance.planetId, prebuildId))
{
   Log.Warn($"BuildFinally was called without having a corresponding PrebuildRequest for the prebuild {prebuildId} on the planet {__instance.planetId}");
   return false;
 }
```

PR #281, opened 2021-05-03
Bug Fix: Prebuilds cannot be build, when host loads game from file

Baldie-dev:
This is a bug fix for #280 .

# Issue
Issue is, that host is checking if existing `PrebuildRequest` exists when `BuildFinally()` is called:
```cs
if (!FactoryManager.ContainsPrebuildRequest(__instance.planetId, prebuildId))
{
Log.Warn($"BuildFinally was called without having a corresponding PrebuildRequest for the prebuild {prebuildId} on the planet {__instance.planetId}");
return false;
}
```

When host loads game from file, this array is empty and all saved prebuilds cannot be build, since it will not pass this check.

# Fix

I have added a method that is executed when the game is loaded and it will go through all existing `prebuilds` and add them to the `PrebuildRequest` array.

### DIFF
--- a/NebulaWorld/Factory/FactoryManager.cs
+++ b/NebulaWorld/Factory/FactoryManager.cs
@@ -4,6 +4,7 @@ using NebulaModel.Logger;
 using NebulaModel.Packets.Factory.Inserter;
 using System.Collections.Generic;
 using UnityEngine;
+using System.Linq;
 
 namespace NebulaWorld.Factory
 {
@@ -41,16 +42,15 @@ namespace NebulaWorld.Factory
             {
                 using (GetPrebuildRequests(out var prebuildRequests))
                 {
-                    for (int i = 0; i < GameMain.data.factoryCount; i++)
-                    {
-                        PlanetFactory factory = GameMain.data.factories[i];
+                    foreach (PlanetFactory factory in GameMain.data.factories)
+                    { 
                         if (factory != null)
                         {
-                            for (int j = 0; j < factory.prebuildPool.Length; j++)
+                            foreach (PrebuildData prebuild in factory.prebuildPool)
                             {
-                                if (factory.prebuildPool[j].id != 0)
+                                if (prebuild.id != 0)
                                 {
-                                    prebuildRequests[new PrebuildOwnerKey(factory.planetId, factory.prebuildPool[j].id)] = LocalPlayer.PlayerId;
+                                    prebuildRequests[new PrebuildOwnerKey(factory.planetId, prebuild.id)] = LocalPlayer.PlayerId;
                                 }
                             }
                         }
--- a/NebulaWorld/Factory/FactoryManager.cs
+++ b/NebulaWorld/Factory/FactoryManager.cs
@@ -46,11 +46,11 @@ namespace NebulaWorld.Factory
                     { 
                         if (factory != null)
                         {
-                            foreach (PrebuildData prebuild in factory.prebuildPool)
+                            for(int i = 0; i < factory.prebuildCursor; i++)
                             {
-                                if (prebuild.id != 0)
+                                if (factory.prebuildPool[i].id != 0)
                                 {
-                                    prebuildRequests[new PrebuildOwnerKey(factory.planetId, prebuild.id)] = LocalPlayer.PlayerId;
+                                    prebuildRequests[new PrebuildOwnerKey(factory.planetId, factory.prebuildPool[i].id)] = LocalPlayer.PlayerId;
                                 }
                             }
                         }
--- a/NebulaWorld/Factory/FactoryManager.cs
+++ b/NebulaWorld/Factory/FactoryManager.cs
@@ -34,6 +34,31 @@ namespace NebulaWorld.Factory
             TargetPlanet = PLANET_NONE;
         }
 
+        public static void InitializePrebuildRequests()
+        {
+            //Load existing prebuilds to the dictionary so it will be ready to build
+            if (LocalPlayer.IsMasterClient)
+            {
+                using (GetPrebuildRequests(out var prebuildRequests))
+                {
+                    for (int i = 0; i < GameMain.data.factoryCount; i++)
+                    {
+                        PlanetFactory factory = GameMain.data.factories[i];
+                        if (factory != null)
+                        {
+                            for (int j = 0; j < factory.prebuildPool.Length; j++)
+                            {
+                                if (factory.prebuildPool[j].id != 0)
+                                {
+                                    prebuildRequests[new PrebuildOwnerKey(factory.planetId, factory.prebuildPool[j].id)] = LocalPlayer.PlayerId;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
         public static void SetPrebuildRequest(int planetId, int prebuildId, ushort playerId)
         {
             using (GetPrebuildRequests(out var prebuildRequests))
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -396,10 +396,7 @@ namespace NebulaWorld
             }
 
             //Initialization on the host side after game is loaded
-            if (LocalPlayer.IsMasterClient)
-            {
-                FactoryManager.InitializePrebuildRequests();
-            }
+            FactoryManager.InitializePrebuildRequests();
 
             LocalPlayer.SetReady();
 
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -395,6 +395,12 @@ namespace NebulaWorld
                 LocalPlayer.Data.Mecha.TechBonuses.UpdateMech(GameMain.mainPlayer.mecha);
             }
 
+            //Initialization on the host side after game is loaded
+            if (LocalPlayer.IsMasterClient)
+            {
+                FactoryManager.InitializePrebuildRequests();
+            }
+
             LocalPlayer.SetReady();
 
             IsGameLoaded = true;


Mushroom:
Just to clarify, the regular drone range stuff for actual building still applies, and if these prebuilts are on another planet the hosts drones won't go and fly off into the unknown?

Baldie-dev:
Yes, currently, drones are going only to the prebuilds that are on the same planet as the player.
This change does not affect drones, it is only affecting the host's list of prebuilds that are ready to build.

This was implemented to prevent calling `BuildFinally()` twice on the same prebuild. So, every time prebuild is created, the host will add it to the array, if prebuild is build it is removed from the array. If `BuildFinally()` is called on prebuild that is not in the array, it will prevent executing `BuildFinally()` method because the prebuild was already build by somebody else.

The issue is, that the array was populated only based on the real-time building events but not based on the loaded save-file. So when the host loaded save-file the array was empty and existing prebuilds were never build.

NebulaWorld/Factory/FactoryManager.cs
```diff
@@ -34,6 +34,31 @@ public static void Initialize()
...
+        public static void InitializePrebuildRequests()
+        {
+            //Load existing prebuilds to the dictionary so it will be ready to build
+            if (LocalPlayer.IsMasterClient)
+            {
+                using (GetPrebuildRequests(out var prebuildRequests))
+                {
+                    for (int i = 0; i < GameMain.data.factoryCount; i++)
+                    {
+                        PlanetFactory factory = GameMain.data.factories[i];
```
Mushroom, line 46: Perhaps a `foreach` with either a linq `where()` or a null check would work better, and the same for below

Mushroom (changes requested):

NebulaWorld/Factory/FactoryManager.cs
Baldie-dev: I was reading some optimization guides for Unity games when I was doing mobile games and at that time the `for` loop was better from the performance point of view than `foreach`. There is some extra overhead or memory handling.

Mushroom: Ahh, okay - than perhaps I will leave this one to @hubastard as the resident Unity expert

Baldie-dev: But the difference is almost 0 in this case of small arrays, so I will change it so it will be easier to read.

Baldie-dev: I just changed for loop to foreach loops :)

NebulaWorld/Factory/FactoryManager.cs
```diff
@@ -34,6 +35,30 @@ public static void Initialize()
...
+        public static void InitializePrebuildRequests()
+        {
+            //Load existing prebuilds to the dictionary so it will be ready to build
+            if (LocalPlayer.IsMasterClient)
+            {
+                using (GetPrebuildRequests(out var prebuildRequests))
+                {
+                    foreach (PlanetFactory factory in GameMain.data.factories)
+                    { 
+                        if (factory != null)
```
Therzok, line 47: Why are there nulls here?

NebulaWorld/Factory/FactoryManager.cs
```diff
@@ -34,6 +35,30 @@ public static void Initialize()
...
+                        {
+                            foreach (PrebuildData prebuild in factory.prebuildPool)
+                            {
+                                if (prebuild.id != 0)
```
Therzok, line 51: Isn't the array tail full with 0's? We could eventually break before iterating everything.

NebulaWorld/SimulatedWorld.cs
```diff
@@ -395,6 +395,12 @@ public static void OnGameLoadCompleted()
                 LocalPlayer.Data.Mecha.TechBonuses.UpdateMech(GameMain.mainPlayer.mecha);
             }
 
+            //Initialization on the host side after game is loaded
+            if (LocalPlayer.IsMasterClient)
```
Therzok, line 399: This looks like a duplicated condition. The logic could simply exist in FactoryManager.

NebulaWorld/Factory/FactoryManager.cs
Baldie-dev: That is true, but I am not sure if it is always in order. For example, the host will build 3 prebuilds and drones will finish the middle one. In that case, there will be:
`[-1, 0, -3, 0, 0, 0, 0]`

NebulaWorld/SimulatedWorld.cs
Baldie-dev: I think in the `FactoryManager` it is not executed after the game fully loads. I think, when I place it in `Initialize()` method, it was executed before the `factory.prebuildPool` was initialized. So I needed to execute this after the game is fully loaded.

NebulaWorld/Factory/FactoryManager.cs
Baldie-dev: I am not sure, it was just "to be safe".

NebulaWorld/Factory/FactoryManager.cs
Therzok: Tl:dr:

`foreach` on `T[]` js the same as `for` loop.

Lambdas are horrible for perf, especially when you capture locals or this.

NebulaWorld/Factory/FactoryManager.cs
Therzok: Doesn't the game compact the array gaps or at least have a counter for the last prebuild ID used?

NebulaWorld/SimulatedWorld.cs
Therzok: Yeah, but the if statement is duplicated.better let FactoryManager do the masterclient check and not force that check into the callers.

NebulaWorld/Factory/FactoryManager.cs
Therzok: LINQ was designed for cases where data is _lazily_ evaluated. It just became common use because of functional programming, but not all LINQ methods allow you to write closure-less callbacks.

In example, `.Where()` on IO or a database might make sense, since that allocation will never cost more than the actual access to the slow resource.

NebulaWorld/Factory/FactoryManager.cs
Therzok: Yup, the game uses `[0, prebuildCursor)` range for values and `prebuild[i].id == i` to check for an instance being valid.

Baldie-dev: I tried this:
```cs
using (GetPrebuildRequests(out var prebuildRequests))
{
    foreach (PlanetFactory factory in GameMain.data.factories)
    { 
        if (factory != null)
        {
            foreach (PrebuildData prebuild in factory.prebuildPool)
            {
                if (prebuild.id == 0)
                {
                    break;
                }
                prebuildRequests[new PrebuildOwnerKey(factory.planetId, prebuild.id)] = LocalPlayer.PlayerId;
            }
        }
    }
}
```
But it did not worked. :( Let me print the content of the array...

Baldie-dev: So, this:
```cs
string s = "";
foreach (PrebuildData prebuild in factory.prebuildPool)
{
    s += prebuild.id + ",";
    if (prebuild.id != 0)
    {
        prebuildRequests[new PrebuildOwnerKey(factory.planetId, prebuild.id)] = LocalPlayer.PlayerId;
    }
}
Debug.Log(s);
```

will produce:
image

NebulaWorld/SimulatedWorld.cs
Baldie-dev: Fixed.

NebulaWorld/Factory/FactoryManager.cs
Therzok: You need to iterate from 0 to prebuildcursor. The list itself does not use prebuildCursor.

Therzok: The iteration should be like

```csharp
for (int i = 0; i < prebuildCursor; i++)
{
    if (prebuildList[i] == i) { /* stuff */ }
}
```

Therzok: preBuildCursor defines the end of the list, not list.Length. might seem like micro-optimizing, but they did it for a reason.

Baldie-dev: fixed :)

NebulaWorld/Factory/FactoryManager.cs
hubastard: For loops are faster than foreach loops in most cases. But it is fine here. I also try to stay away of LINQ in unity. Since it is slower and of habits since it doesn't work well on iOS because it needs a JIT compiler.

Also, fun fact, in Unity because of .net 3.5 it is faster to do `++i` instead of `i++` in a for loop

hubastard (approved):

Therzok (approved):

NebulaWorld/Factory/FactoryManager.cs
Therzok: Array iteration is identical for loops and foreach loops. If the perf differs, it's because the dll was built with `mcs` - mono C# compiler - not `csc` (aka roslyn).

Therzok: List and other collections are different.

Therzok: https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE+UkALIgLIAUAlgHYAuA2gLqICGATi0wJ4CUm2u/AMwD2LAKZMAxgAtEpAG6tmbRFSXtuvGP2060QygGcArgFtRpViy4BuPjoC+dxI624nCZETJCARgCtRCTpGS04eLFcdYTFJGXlFSxVKNTDNXXS9Q1NzSxsnfhcCp3d8IgAxEQoaBhSNCPTo2SpqFUQAXkQABmtWgB5agDoAGVFKAHNqKR6AamnycPzdVH1jMws2TlpyejzI7ULcA5xivcJM1dFegBUAPlIrxAAPLgwXeyA===

Issue #285, opened 2021-05-04
Multiplayer Options Float Sliders Don't Have Proper Stepsize


PhantomGamers:
When adding a slider to a float config option, the logic of the step size should be:
`(!wholeNumbers) ? ((maxValue - minValue) * 0.1f) : 1f`

However, it's set to 1f regardless of what `wholeNumbers` is set to in the UIOptionWindow_Patch class.

I thought this might be because of how the slider is created before the patch assigns the values, so it wasn't considering that for the step size, but it looks like the values I'm assigning are the default values in the actual Slider class so it should have a step size of 0.1f.

hubastard:
I'm looking at it right now

Issue #143, opened 2021-04-14
Possibility to change nickname


Baldie-dev:
Currently, players are having nicknames in the format `Player x` where `x` is player id. Maybe it would be nice to have UI for players to be able to choose their own nickname.

PR #286, opened 2021-05-04
Add support for configuring nickname and color

### LINKED ISSUE #142: Possibility for players to change color in game
Currently, players are having random colors, maybe it would be great to add any UI for players to be able to change the color.

PhantomGamers:
Color is currently configured with separate RGB text boxes.

I was going to enable the slider but can't because of #285 

Adding Float3 support to the menu with an actual color picker would be ideal but this works for now.

Closes #142 #143


NebulaModel/MultiplayerOptions.cs
```diff
@@ -7,6 +7,21 @@ namespace NebulaModel
     [System.Serializable]
     public class MultiplayerOptions : ICloneable
     {
+        [DisplayName("Nickname")]
+        public string Nickname { get; set; } = "Player";
```
hubastard, line 11: We could use `AccountData.me.userName` as default value

NebulaModel/MultiplayerOptions.cs
```diff
@@ -7,6 +7,21 @@ namespace NebulaModel
...
+        [UIRange(0f, 1f)]
+        public float ColorR { get; set; } = 1f;
+
+        [DisplayName("Color Green")]
+        [UIRange(0f, 1f)]
+        public float ColorG { get; set; } = 0.6846404f;
+
+        [DisplayName("Color Blue")]
+        [UIRange(0f, 1f)]
+        public float ColorB { get; set; } = 0.24313718f;
```
hubastard, line 23: We should use the name `Mecha Color Red` to be more explicit

NebulaModel/MultiplayerOptions.cs
PhantomGamers: I wanted to do this but the config generates before AccountData.me.userName is populated

NebulaModel/MultiplayerOptions.cs
PhantomGamers: > We should use the name `Mecha Color Red` to be more explicit

Should I reflect this in the variable name as well?

NebulaModel/MultiplayerOptions.cs
Shad0wlife: Do you know when that value gets populated and then auto-set this value to it? Or is that not viable (performance, or the population is too late)?

NebulaModel/MultiplayerOptions.cs
hubastard: yes please

PhantomGamers: > yes please

Done in 58edede

NebulaModel/MultiplayerOptions.cs
PhantomGamers: I could keep the default null and make a postfix patch for PARTNER.Login() and do Config.Options.Nickname = AccountData.me.userName ?? "Player"; there I suppose.

Would this be desireable?

Some games, though usually older ones, with official multiplayer support default to just calling the player "Player" so I figured this was just the cleanest approach but it's up to you folks.

Therzok (approved):

PR #296, opened 2021-05-13
add tutorial and advisor disable settings in MultiplayerOptions

NoGTNoHappy:


NoGTNoHappy:
close #254

hubastard (approved):

NebulaModel/MultiplayerOptions.cs
```diff
@@ -11,6 +11,12 @@ public class MultiplayerOptions : ICloneable
         [UIRange(1, ushort.MaxValue)]
         public ushort HostPort { get; set; } = 8469;
 
+        [DisplayName("Disable Tutorials")]
+        public bool TutorialDisabled { get; set; } = true;
```
Therzok, line 15: Shoudl they be disabled by default?

NoGTNoHappy: I thought players using nebula may not want them by default cuz they may not be new to DSP, and these fuctions don't effect single player mode.

Issue #289, opened 2021-05-06
Researchdesync research finishing when multiple techs are in queue


pGlase:
This is a follow-up to #244 , was tested on newest build from server artifact of #288 .
Steps to reproduce:
-Server queues n(>=3) research tech
-Client joins -> correct sync
-Research finishes -> tech no1. unlocks for server and client
-> Client dequeues the research no.2 and starts progress on tech no.3
-> Server starts progress on tech no.2

~~I have a hunch that the onTechUnlocked-method is trying to dequeue the current research.~~
Wrong assumption.
As we force the client in NebluarClient.PacketProcessors.GameHistory.GameHistoryUnlockTechProcessor to dequeue tech, this might be too much.

If nobody is interested in this, i can look into this over the weekend.

avalanchas:
Do you like more help on reproducing this? I have no clue of mod programming or even C# coding in general, but we faced this issue a lot today in a 3-person multiplayer game, and I think I can reliably reproduce glitches with the research mechanic

We several issues:
- queue not showing the same for all players
- queue taking resources (early game, before cubes) from one players inventory and then not accepting other type of resource from other player
- research success message visible for all players, but tech not actually researched (buildings not available in menu) unless the server host enqueues the research

I will try newest build from pipeline tomorrow and let you know which of these persist

pGlase:
I'm fine on the testing for now, but I couldn't invest much time on this as some things in RL need my attention right now.
And while I like your enthusiasm, I need you to slow you down, as this issue is only intended for the research desync on finishing tech research, not all your points 😅.

I'll give you a quick opinion on your points:

-Queue not showing the same for all players
That's possibly the desync issue here, but I need more info on what you guys did before the desync happened in order to help you 😄.
If it's not on finishing research, please open a new issue with steps on how to reproduce it.

-Queue taking resources [...]
I think you mean that Player A wants to only contibute Item A and Player B only wants to contribute Item B for a tech, but the research is not starting and the enqueueing player "looses" their Items.

I've tried my hand on this in a previous defect, but couldn't get something reasonable to work, as the game is "hard-wired" to only progress the current research if at least one item of all needed types are avaliable at the same time to the (single)player.
To fix this for the MP, I'd need to sync and refund a lot of items, which didn't turn out "nice", so grudgingly left it in it's current state.
I'd be really interested to hear more if the enqueuing player doesn't get their items back after cancelling the research.

-Research success message visible for all players, but tech not actually researched  [...]
[...]

PR #299, opened 2021-05-17
fix for #289

pGlase:
Changed unlockorder to first dequeue researched tech and then unlock it for the client.
This prevents dequeuing of the next item in the queue  for the non-researching players, which lead to desyncing.


sp00ktober (approved):
LGTM, did not have the time to test yet but i dont think it‘ll break something 😁

Therzok (approved):

hubastard (approved):

Issue #276, opened 2021-05-03
Other mechas seem to keep the graphic for "warping" and displaying it for everyone else


Jon-Myrstad:
Seems somewhat random wether it gets stuck in this state or not.

![image](https://user-images.githubusercontent.com/7058316/116832149-ec7c5900-abb3-11eb-8871-f4d250bb3350.png)

sp00ktober:
I have run into this too couple of times, not sure why its happening. That animation should be triggered by the player speed, so if he stands still or even when he sails this should only happen at a specific speed limit.

Any ways i need to look into this, thanks.

sp00ktober:
Seems to be caused when a player arrives at a planet and needs to receive the factory from server resulting in a small lag and jump after loading it. This jump triggers the warping animation and it seems to get stuck in there then.

I will change this and add a small packet that tells other players when the local player goes in/out of warping state to decouple this from the movement.

PhantomGamers:
This can be closed as it will be fixed when the PR for 0.7.x is merged

PR #304, opened 2021-05-25
fix for #276 refactor of warp animation trigger

sp00ktober:
The warp animation is no longer triggered by player movement speed (which caused #276 most likely due to the short freezes when loading a planet's factory from the server and the followed "jump").

It is now triggered by a network packet that gets send once the player truly enters/exits warping state.

### DIFF
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -105,10 +105,8 @@ namespace NebulaWorld
         {
             using (GetRemotePlayersModels(out var remotePlayersModels))
             {
-                Debug.Log("trying to add " + playerData.PlayerId);
                 if (!remotePlayersModels.ContainsKey(playerData.PlayerId))
                 {
-                    Debug.Log("added");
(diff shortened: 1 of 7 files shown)


sp00ktober:
i pushed this to https://github.com/hubastard/nebula/pull/307 so it will be included in the version for 0.7.x (but still would be cool to have this for 0.6.x until we are done)

NebulaWorld/SimulatedWorld.cs
```diff
@@ -150,6 +150,25 @@ public static void UpdateRemotePlayerAnimation(PlayerAnimationUpdate packet)
             }
         }
 
+        public static void UpdateRemotePlayerWarpState(PlayerUseWarper packet)
+        {
+            using (GetRemotePlayersModels(out var remotePlayersModels))
+            {
+                if(packet.PlayerId == 0) packet.PlayerId = 1; // host sends himself as PlayerId 0 but clients see him as id 1
```
hubastard, line 157: We should fix this instead if its the case. We might have some unknown bugs related to this issue.

sp00ktober:
This is fixed in the current code, so this pr can be closed.

PR #366, opened 2021-07-10
Configuration file refactor  to use BepInEx config file system

hubastard:
## Breaking Changes
- Configuration file refactor  to use BepInEx config file system. 
    - **Please note that this will reset your current multiplayer settings.**

### DIFF
--- a/NebulaModel/Config.cs
+++ b/NebulaModel/Config.cs
@@ -1,67 +1,78 @@
 ﻿using BepInEx;
+using BepInEx.Configuration;
+using HarmonyLib;
 using NebulaModel.Logger;
 using System;
+using System.Collections.Generic;
 using System.IO;
-using System.Runtime.Serialization.Formatters.Binary;
+using System.Linq;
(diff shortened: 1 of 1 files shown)


NebulaModel/Config.cs
```diff
@@ -1,67 +1,77 @@
...
-                try
-                {
-                    var formatter = new BinaryFormatter();
-                    using (Stream stream = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.None))
-                    {
-                        Options = (MultiplayerOptions)formatter.Deserialize(stream);
-                    }
-                }
-                catch(System.Exception e)
+                configFile = new ConfigFile(Path.Combine(Paths.ConfigPath, OPTION_SAVE_FILE), true);
```
PhantomGamers, line 30: We don't really need to specify the path, it's probably best to just use the default config location? (`BepInEx\config`)

sp00ktober (approved):

PR #369, opened 2021-07-11
Update build-winx64.yml

PhantomGamers:
* Ensure artifact zips begin with "Nebula\" in the root
* Include SHA7 in artifact names
* Update upload-artifact to v2.2.4

### DIFF
--- a/.github/workflows/build-winx64.yml
+++ b/.github/workflows/build-winx64.yml
@@ -19,6 +19,7 @@ jobs:
 
     env:
       Solution_Name: Nebula.sln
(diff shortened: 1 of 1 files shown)


PhantomGamers:
Perhaps before merging this we should also prep for the Thunderstore release by including the icon.png, README.md, and manifest.json within the Nebula/ folder as described on https://thunderstore.io/package/create/

r2modman (the mod manager for Thunderstore) installs the mods to `BepInEx\plugins\{author}-{modname}` so in that case we should follow the same rules when building and packaging the mod here.

PhantomGamers:
closing this as these changes will be integrated in some other pr (#399 or #408 probably)

PR #371, opened 2021-07-12
Improve remote player ui visuals

Slappy826:
- In-game name will resize automatically based on distance to the camera to prevent the text from being too big up close
- In-game name no longer has blurry text
- Star map players now match the tracker for the local player

[Video preview of changes](https://youtu.be/j6lO9syl0BA)

### DIFF
--- a/NebulaWorld/SimulatedWorld.cs
+++ b/NebulaWorld/SimulatedWorld.cs
@@ -469,6 +469,7 @@ namespace NebulaWorld
         {
             // Make a copy of the "Icarus" text from the starmap
             Text starmap_playerNameText = (Text)AccessTools.Field(typeof(UIStarmap), "playerNameText").GetValue(starmap);
+            Transform starmap_playerTrack = (Transform)AccessTools.Field(typeof(UIStarmap), "playerTrack").GetValue(starmap);
 
             using (GetRemotePlayersModels(out var remotePlayersModels))
             {
@@ -477,11 +478,11 @@ namespace NebulaWorld
                     RemotePlayerModel playerModel = player.Value;
 
                     Text nameText;
-                    Text starmapMarker;
-                    if (playerModel.StarmapNameText != null && playerModel.StarmapMarker != null)
+                    Transform starmapTracker;
+                    if (playerModel.StarmapNameText != null && playerModel.StarmapTracker != null)
                     {
                         nameText = playerModel.StarmapNameText;
-                        starmapMarker = playerModel.StarmapMarker;
+                        starmapTracker = playerModel.StarmapTracker;
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


hubastard (approved):

NebulaWorld/SimulatedWorld.cs
```diff
@@ -524,12 +523,27 @@ public static void RenderPlayerNameTagsOnStarmap(UIStarmap starmap)
...
+                        starmapTracker.rotation = rotation;
+                    }
+
+                    starmapTracker.localScale = UIStarmap.isChangingToMilkyWay ? Vector3.zero : 
+                        Vector3.one * (starmap.screenCamera.transform.position - starmapTracker.position).magnitude;
 
                     // Put their name above or below it
-                    nameText.rectTransform.anchoredPosition = new Vector2(rectPoint.x, rectPoint.y + (rectPoint.y >= -350.0 ? -19f : 19f));
-                    nameText.transform.localScale = Vector3.one;
+                    nameText.rectTransform.anchoredPosition = new Vector2(rectPoint.x + (rectPoint.x > 600f ? -35 : 35), rectPoint.y + (rectPoint.y > -350.0 ? -19f : 19f));
```
Mushroom, line 545: Have you tested this at the minimum resolution? I found some interesting behaviour with it, hence me picking the values I did when I wrote this

NebulaWorld/SimulatedWorld.cs
```diff
@@ -607,6 +619,26 @@ public static void RenderPlayerNameTagsInGame()
 
                     // Make sure the text is pointing at the camera
                     playerNameText.transform.rotation = GameCamera.main.transform.rotation;
+
+                    // Resizes the text based on distance from camera for better visual quality
+                    var distanceFromCamera = Vector3.Distance(playerNameText.transform.position, GameCamera.main.transform.position);
+                    var nameTextMesh = playerNameText.GetComponent<TextMesh>();
+
+                    if (distanceFromCamera > 100f)
```
Mushroom, line 627: IMO some kind of linear interpolation would look better than the jumpy steps that happen here

Mushroom (changes requested):
Thanks for this, I was going to come back around and look at this again at some point, the current implementation was very much a case of "Get it working first, make it pretty later".

Untested, my comments are based on your video. I will not be able to test in person until Sunday afternoon/evening, so please post another video if you make some changes so I can judge off of that.

Also, what does this look like in the galaxy view?

NebulaWorld/SimulatedWorld.cs
Slappy826: I changed it to match what the game itself does, so there wouldn't be differing behavior. (although the game also rounds the values, but i didn't implement that.)

sp00ktober:
Is this ready for merge?

NebulaWorld/SimulatedWorld.cs
PhantomGamers: Something like this?

```cs
nameTextMesh.characterSize = Mathf.Lerp(0.1f, 0.2f, Mathf.Clamp(distanceFromCamera / 100, 0, 1));
nameTextMesh.fontSize = (int)Mathf.Lerp(36, 60, Mathf.Clamp(distanceFromCamera / 100, 0, 1));
```

PR #373, opened 2021-07-14
fix for the Power Tower error

sp00ktober:
As the game is threaded now and the PowerTower transpiler code seemingly runs inside them, it accessed a not thread save Dictionary potentially resulting in that error.

Now using a thread save Dictionary should solve it.

### DIFF
--- a/NebulaWorld/Factory/PowerTowerManager.cs
+++ b/NebulaWorld/Factory/PowerTowerManager.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
 
 namespace NebulaWorld.Factory
 {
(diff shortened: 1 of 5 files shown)


NebulaWorld/Factory/PowerTowerManager.cs
```diff
@@ -225,7 +224,11 @@ public static void AddExtraDemand(int PlanetId, int NetId, int NodeId, int Power
                 map.ExtraPower = PowerAmount;
 
                 mapping2.Add(map);
-                Energy.Add(PlanetId, mapping2);
+                if (!Energy.TryAdd(PlanetId, mapping2))
+                {
+                    // if we failed to add then most likely because another thread was faster, so call this again to run the above part of the method.
```
PhantomGamers, line 229: Is it possible for this to keep failing indefinitely? Should there be a condition where it stops trying?

sp00ktober: I dont think so, this only happens if some player walks into the range of a power tower, and then it calls it once per tower.
I think this is only relevant anyways if you walk into the range for more than one tower at exactly the same time

PhantomGamers (approved):

sp00ktober:
i need to add one more fix to it before merge

PR #374, opened 2021-07-14
Fix lag/freezes caused by #367

PhantomGamers:
We don't need to initialize planet physics and audio for destruct entity requests.


sp00ktober (approved):
lgtm

PR #377, opened 2021-07-18
Refactor loading of planet data

PhantomGamers:
Rather than load and unload planet data (physics, audio, etc.) for every request when the host is not on the same planet that the request is coming from, let's load it once and only unload it if we have gone some time without a request.

I have set this timeout period to 10 seconds but perhaps another value would be preferred, or maybe it should be user configurable?

### DIFF
--- a/NebulaWorld/Factory/FactoryManager.cs
+++ b/NebulaWorld/Factory/FactoryManager.cs
@@ -1,8 +1,11 @@
-﻿using HarmonyLib;
+﻿using BepInEx;
+using HarmonyLib;
 using NebulaModel.DataStructures;
 using NebulaModel.Logger;
 using NebulaModel.Packets.Factory.Inserter;
+using System;
 using System.Collections.Generic;
+using System.Timers;
 using UnityEngine;
 
 namespace NebulaWorld.Factory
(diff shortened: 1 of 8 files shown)


sp00ktober:
Does this mean the host is unable to place buildings for that time on his planet?

PhantomGamers:
> Does this mean the host is unable to place buildings for that time on his planet?

No, this is purely about the loading of planet data such as physics, audio, and render batches. This does mean that while the planet is loaded the host will be using more resources, since they will have multiple planets loaded at once. If it's just one additional planet it's not bad, but it might be an issue with many players all on different planets, though even in that case I think it should be better than it was before if they are all building at the same time anyway

sp00ktober (approved):
Looks good to me

NebulaWorld/Factory/FactoryManager.cs
```diff
@@ -34,6 +41,114 @@ public static void Initialize()
...
+                }
+
+                // .NET Framework 4.7.2 does not have TryAdd so we must make sure the dictionary does not already contain the planet
+                if (!planetTimers.ContainsKey(planetId))
+                {
+                    // We haven't loaded this planet, let's load it
+                    LoadPlanetData(planetId);
+
+                    // Create a new 10 second timer for this planet
+                    planetTimers.Add(planetId, new Timer(TimeSpan.FromSeconds(10).TotalMilliseconds));
```
Therzok, line 61: What stops the timers? This timer should be AutoReset=false or something otherwise this will keep triggering every 10 seconds and uses a threadpool thread to do it.

PhantomGamers: When the timer elapses it calls PlanetTimer_Elapsed(int planetId) which calls RemovePlanetTimer(planetId) which stops the timer and sets it to null

Therzok: Setting it to null won't stop the timer :p

Therzok: > This type implements the IDisposable interface. When you have finished using the type, you should dispose of it either directly or indirectly. To dispose of the type directly, call its Dispose method in a try/catch block. To dispose of it indirectly, use a language construct such as using (in C#) or Using (in Visual Basic). For more information, see the "Using an Object that Implements IDisposable" section in the IDisposable interface topic.

Therzok: It's possible that the Timer implementation uses weak keys to store the timer, but Stop only stops the timer from being fired, not from being added to the global timer queue.

Therzok: Oh, it's wrapped in a using statement there, I missed it in my initial review. It's fine.

Therzok (approved):

NebulaWorld/Factory/FactoryManager.cs
PhantomGamers: Hrm so the using statement is wrapping the Locker GetPlanetTimers() object, would the timer itself also be disposed after this? I copied the GetPrebuildRequests implementation when I did this so I don't fully understand how it works tbh, maybe disposal is required 😅

Issue #378, opened 2021-07-19
Host can generate NRE when loading a planet that a client is building on


PhantomGamers:
as in, when the host is arriving on the planet

![image](https://user-images.githubusercontent.com/844685/126113151-0cc6f8d8-a039-4a95-a10a-5d6fd96f67b4.png)

This might already be fixed with the currently method we are using in master, but the problem with the method being used now is that clients can receive multiple items back when they are dismantling items due to PlayerAction_Build.DoDismantleObject() potentially running multiple times

PhantomGamers:
With #377 this can be triggered from the LoadPlanetData function calling CreateRenderingBatches() hmm
![image](https://user-images.githubusercontent.com/844685/126589950-b8c33bc9-2785-4a96-b719-2f3489539a58.png)

PR #379, opened 2021-07-19
Refactor UpgradeEntityRequests and DestructEntityRequests

PhantomGamers:
PlayerAction_Build.Do(Destruct/Upgrade)Object() was being run multiple times resulting in players receiving multiple of the same item so we should ensure this only runs once

This may bring back #378, needs further testing, but fixes the other problem mentioned on that issue.


PhantomGamers:
Also just to add, I wasn't able to reproduce #378 locally with this PR, but it's been notoriously hard to reproduce alone while trying to control 2 clients so it *could* still be an issue

PhantomGamers:
I'm gonna merge this because no one has reported any errors so far and if anything this definitely fixes an item dupe bug which should be included in the next release

hubastard (approved):

PR #386, opened 2021-07-22
fix newly introduced multithread issue by locking access

sp00ktober:
regarding the PowerTowers.


PhantomGamers:
LGTM but maybe it would be better to use NebulaModel.DataStructures.Locker for this?

PhantomGamers (approved):

hubastard (approved):

PR #393, opened 2021-07-24
NameInputPacket: Fix compatibility with GS

PhantomGamers:
Assuming ids of over 100 are planets doesn't work when there are 1000 stars, so let's just send either a planetid or a starid so we can determine which it is

### DIFF
--- a/NebulaClient/PacketProcessors/Universe/NameInputProcessor.cs
+++ b/NebulaClient/PacketProcessors/Universe/NameInputProcessor.cs
@@ -22,18 +22,17 @@ namespace NebulaClient.PacketProcessors.Universe
             }
             using (FactoryManager.EventFromServer.On())
             {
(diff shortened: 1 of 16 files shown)


NebulaClient/PacketProcessors/Universe/NameInputProcessor.cs
```diff
@@ -22,18 +22,17 @@ public void ProcessPacket(NameInputPacket packet, NebulaConnection conn)
             }
             using (FactoryManager.EventFromServer.On())
             {
-                // If stellarId > 100 then it must be a planet
-                if (packet.StellarId > 100)
+                if (packet.StarId != -1)
```
Therzok, line 25: Maybe introduce a constant field for StarId == -1, would explain what the ID is.

hubastard (approved):

PR #394, opened 2021-07-24
[Refactor] Packet Processors

hubastard:
This PR is huge because it is a major refactor, I decided to merge the Host and Client packet processor logic. This is the first step in what I have planned to refactor.

Please note that some packet processor were merged quickly which trying to optimize the codebase (especially for the Logistic Station processors)

## Changes

- Now the `NebulaHost` and `NebulaClient` project are merged into a single project named `NebulaNetwork`
- We now have only 1 `PacketProcessor` class per packet type. It handles both the Host and the Client logic.
- We now have access to 2 new property inside a PacketProcessor: `IsHost` and `IsClient`

Packet Processor now looks like this:

PacketProcessor


PhantomGamers (approved):
other than the ILS stuff (which @sp00ktober would need to look at), the rest of it looks good to me

hubastard:
Ok, I will merge this right now, since I kept the ILS stuff "as is" and only added some TODO where the previous code seemed fishy to me. We will be able to look at it later I change the code if there was really an issue in the ILS code.

Issue #370, opened 2021-07-12
Client's Detail Display settings don't persist between sessions


PhantomGamers:
e.g. power grid & veins distribution visibility

PR #397, opened 2021-07-24
Client's Detail Display settings (e.g. power grid enabled) will now be saved and load from configuration

Needix:
Extended the configuration to save the data.
Patched  UIGameMenu to save the data.
Patched the necessary classes to load the data

Issue https://github.com/hubastard/nebula/issues/370


PhantomGamers (changes requested):
All of the patches should begin with

```
            if (!SimulatedWorld.Initialized || LocalPlayer.IsMasterClient)  
            {  
                return;  
            }
```

so that normal behavior occurs when not in multiplayer and for the host.

NebulaPatcher/Patches/Dynamic/UIDetailsView_Patch.cs
```diff
@@ -0,0 +1,92 @@
...
+using NebulaModel;
+using NebulaModel.Logger;
+using NebulaModel.Packets.Universe;
+using NebulaWorld;
+using NebulaWorld.Factory;
+using UnityEngine.UI;
+
+namespace NebulaPatcher.Patches.Dynamic
+{
+    [HarmonyPatch(typeof(UIGameMenu))]
```
PhantomGamers, line 11: Should this be typeof(UIDetailsView) or should the patch be renamed?

PhantomGamers (changes requested):

PhantomGamers (approved):

PR #399, opened 2021-07-25
Enable automatic Thunderstore releases

PhantomGamers:
Todo:
* ~~Automate releases to thunderstore (in progress)~~


PhantomGamers:
We need to wait for Thunderstore's upload rework pr to be completed and merged before a token can be generated for us to use, so a 0.3.1 release should be made manually in the mean time.

PhantomGamers:
Steps for publishing a build to both github and thunderstore:

- Edit version.json with the desired version if the version that it has been automatically incremented to is not desired (e.g. if you wanted to bump the minor version instead of the build version)
- Edit CHANGELOG.md so that the topmost version is the version that is in version.json, if this is not the case the workflow will fail. Put the changes for the build each on their own line beginning with a -
- Run the Publish Release workflow!
- Get a snack
- High five yourself for the easy release! If all went well the build should now be published to both thunderstore and github. (Thunderstore uploads will only be automated after a TSTORE_TOKEN var is added to the github secrets on the repo)

hubastard (approved):

PR #408, opened 2021-07-30
Switch to using publicized stubs as references

PhantomGamers:
This will allow us to:

* See private members in code directly.
* Get notified by the IDE when a name is changed after an update.
* Allow CI instances that do not have the game installed to build Nebula (so that we can use the Github CIs)

Included is a .github\scripts\stub.js file that when ran from the project root will generate all of the references listed in references.txt. ~~[MrPurple6411's latest AssemblyPublicizer.exe (from the repo itself, not the release!)](https://github.com/MrPurple6411/AssemblyPublicizer/blob/master/AssemblyPublicizer.exe) needs to be extracted to Libs/AssemblyPublicizer/~~

TODO:

* ~~Parse DevEnv.targets if it exists and use that for the DSP directory path.~~
* ~~Fix PlanetData.onLoaded ambiguity error in PlanetDataRequestProcessor.cs~~
* Unify references.txt with their inclusion in Directory.Build.props so that references only need to be added to one location.
* ~~Have stub script automatically download AssemblyPublicizer if needed (or we can include it in the repo maybe? would need to look into the license)~~


PhantomGamers:
I fixed the ambiguity errors by switching to [MrPurple6411's version of AssemblyPublicizer](https://github.com/MrPurple6411/AssemblyPublicizer/blob/master/AssemblyPublicizer.exe) (which has been reflected in the link above as well)

PhantomGamers:
I just realized, that version of AssemblyPublicizer also stubs so we don't need to use Refasmer

PhantomGamers:
This should be in a good state to merge. I didn't unify the references.txt with the directory.build.props because I wasn't sure of exactly how that should be done, but it's not a big deal as we likely aren't going to be removing or adding references often

.github/scripts/stub.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (existsSync(targetsFile)) {
+    const xml = XmlReader.parseSync(readFileSync(targetsFile, "utf-8"));
+    const tmpPath = xmlQuery(xml).find("DSPGameDir").text();
+    if (existsSync(tmpPath)) {
+      nebulaPath = tmpPath;
+    }
+  }
+
+  return join(nebulaPath);
```
hubastard, line 36: Why do you do a `path.join` with only one value ?

hubastard (commented):
What is expose in those .dll ?

hubastard (approved):

.github/scripts/stub.js
PhantomGamers: > Why do you do a `path.join` with only one value ?

it's the same purpose as EnsureTrailingSlash in the directory.build.props, just to make sure it ends in a /... I think it works like that at least 🤔

PhantomGamers:
> 
> 
> What is expose in those .dll ?

They just contain the bare minimum to access the api, so like function names, property names, etc., with none of the actual code so that it is OK to redistribute

PR #409, opened 2021-07-30
Update SyncCompleteProcessor.cs

PhantomGamers:
I had accidentally done it so we  were sending the planet id as the star id argument, whoops.

### DIFF
--- a/NebulaNetwork/PacketProcessors/Session/SyncCompleteProcessor.cs
+++ b/NebulaNetwork/PacketProcessors/Session/SyncCompleteProcessor.cs
@@ -16,7 +16,7 @@ namespace NebulaNetwork.PacketProcessors.Session
 
         public SyncCompleteProcessor()
         {
-            playerManager = MultiplayerHostSession.Instance?.PlayerManager;
+            playerManager = MultiplayerHostSession.Instance != null ? MultiplayerHostSession.Instance.PlayerManager : null;
         }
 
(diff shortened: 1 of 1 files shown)


NebulaNetwork/PacketProcessors/Session/SyncCompleteProcessor.cs
```diff
@@ -16,7 +16,7 @@ public class SyncCompleteProcessor : PacketProcessor<SyncComplete>
 
         public SyncCompleteProcessor()
         {
-            playerManager = MultiplayerHostSession.Instance?.PlayerManager;
```
hubastard, line 19: Isn't this doing the same as it was previously?

PhantomGamers: VS warns against using null propagation on unity objects as they can be invalid without being null

PhantomGamers: https://blog.unity.com/technology/custom-operator-should-we-keep-it

hubastard (approved):

NebulaNetwork/PacketProcessors/Session/SyncCompleteProcessor.cs
Therzok: Yeah, `?.` on anything defined in unity is dangerous.

PR #410, opened 2021-07-31
Switch to Mirror networking library

PhantomGamers:
Tbh maybe this is more trouble than it's worth, we could honestly just modify what we currently have to support slotting in different low level networking options but I figured I would mess around and see if I could get it to work.

Currently, it seems that hosting works (portchecker shows the port as open with the server running), however the client gets stuck at connecting and never makes a connection.

This is only an initial hacky implementation simply to test to see if we can get a connection to be made at all, if that works out we can probably remove a lot of our own packet and player management stuff in favor of the Mirror apis


PhantomGamers:
* We can now communicate between server and client over the Mirror networking system

* Client cannot currently successfully handshake, for some reason `using (playerManager.GetPendingPlayers(out var pendingPlayers))` in NebulaNetwork.HandshakeRequestProcessor.cs is resulting in an NRE

PR #413, opened 2021-08-04
PlanetFactory.PickFrom() surpression and helper for user

sp00ktober:
WIP so i mark it as draft for now.

This surpresses the error spam originated from `PlanetFactory.PickFrom()` due to faulty sorters which can happen by some unknown desync issue and persist in the savegame.

It also adds a text similar to the usernames below the faulty sorters so users can find and replace them.

I need to do a bit of code cleanup and maybe add more checks to find even more faulty sorters (if there are any)


NebulaPatcher/Patches/Dynamic/FactorySystem_Patch.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            if (!SimulatedWorld.Initialized || !LocalPlayer.IsMasterClient)
+            {
+                return true;
+            }
+
+            foreach (var obj in InserterComponent_Transpiler.FaultySortersText)
+            {
+                if (obj.Value?.transform.position == __instance.inserterPool[id].pos2)
+                {
```
Therzok, line 22: Don't use null conditional with unity types. It breaks at runtime since they override ==

PR #416, opened 2021-08-12
Fixes for update 7962

PhantomGamers:
Currently this fixes building and playing, however clients can only see planets/planetary systems in the production window that they visit in the session.


hubastard (approved):

PhantomGamers:
Apparently that issue also exists on 0.3.1 so I guess this can be merged now

PR #426, opened 2021-08-25
Refactor session architecture

hubastard:
This is the current draft PR for the refactor to the new session architecture.

- Basically, everything is now accessible through the `Multiplayer` class.
- The class contains a `Session` property which include all the managers and data related to the current multiplayer session.
- The session data is cleared when the player leave the multiplayer game.

Here are the main changes:
- `SimulatedWorld.Initialized` -> `Multiplayer.IsActive`
- `MultiplayerClientSession.Connect()` -> `Multiplayer.JoinGame()`
- `MultiplayerHostSession.StartServer()` -> `Multiplayer.HostGame()`
- `LocalPlayer` -> `Multiplayer.Session.LocalPlayer`
- `FactoryManager` -> `Multiplayer.Session.Factories`
- `MultiplayerClientSession` has been replaced by the `Client` class.
-  `MultiplayerHostSession` has been replaced by the `Server` class.
- `LocalPlayer.PlayerId` -> `Multiplayer.Session.LocalPlayer.Id`
- `LocalPlayer.IsMasterClient` -> `Multiplayer.Session.LocalPlayer.IsHost`

Before when we wanted to send a packet we were able to do it either using the `LocalPlayer`, `MultiplayerClientSession` or `MultiplayerHostSession`. Now to send a packet you do: `Multiplayer.Session.Network.SendPacket()`

What still needs to be done:
- Right now, before I can test this, we need to fix the Transpiler that was doing for exemple: `DroneManager.BroadcastDroneOrder()` to do `Multiplayer.Session.Drones.BroadcastDroneOrder()`. I'm not really good with transpilers so if @PhantomGamers or @sp00ktober can help me for this that would be great ❤️ 
- I also need to fix the client disconnection flow to work with the new session architecture 

Once the transpilers fix are done, we will be able to test to make sure that everything still works.


PhantomGamers:
I fixed the transpilers, and there were a bunch of instances where Multiplayer.Session could be accessed before it is set that I fixed in 1f7e5a4 by just null propagating it, not sure if this is the best solution.

Currently the host can get in-game but the client is unable to connect, nothing happens after the client registers the packet processors. I'm not sure what is causing this.

hubastard:
Oh right, you should remove the null check of the session in the packet processors that you added because it should never be null at that point. The real fix is to register the packet processors on Client.Start() and Server.Start() instead of in their base class constructor.

Right now since it is in the constructor it means that I was registering the processors before the Session property was set, hence the null ref in the packet processors.

hubastard:
And now apart from that packet processor issue, we should never need to do a null check on the Session. If we have a NRE on the session somewhere, it means that we are missing a Multiplayer.isActive check.

NebulaNetwork/PlayerManager.cs
```diff
@@ -214,24 +213,15 @@ public void SendPacketToOtherPlayers<T>(T packet, Player sender) where T : class
...
-            PlayerData playerData;
-            if (LocalPlayer.GS2_GSSettings != null)
-            {
-                playerData = new PlayerData(playerId, -1, PlayerColor, position: new Double3(birthPlanet.uPosition.x, birthPlanet.uPosition.y, birthPlanet.uPosition.z));
-            }
-            else
-            {
-                playerData = new PlayerData(playerId, -1, PlayerColor);
-            }
+            PlayerData playerData = new PlayerData(playerId, -1, PlayerColor);
```
kremnev8, line 222: This will break compatibility with Galactic Scale 2. In GS2 birth system can be not at zero. Because of that you need to set position. Without GS2 it should default to (0,0,0)
```cs
Float3 PlayerColor = new Float3(Config.Options.MechaColorR / 255, Config.Options.MechaColorG / 255, Config.Options.MechaColorB / 255);
PlanetData birthPlanet = GameMain.galaxy.PlanetById(GameMain.galaxy.birthPlanetId);
PlayerData playerData = new PlayerData(playerId, -1, PlayerColor, position: new Double3(birthPlanet.uPosition.x, birthPlanet.uPosition.y, birthPlanet.uPosition.z));
```

PhantomGamers: Done in cc2ff2f

Issue #435, opened 2021-09-08
Auto-pause when no clients are connected


jjcm:
Have a way to auto-pause when no clients are connected. I'd love to be able to get a dedicated server going that only runs when we've joined remotely.

PhantomGamers:
closed thanks to @starfi5h !

PR #438, opened 2021-09-10
add CanPause check, enable pausing when no connections

starfi5h:
Attempt to implement #435
- Add `Multiplayer.CanPause` (not sure if it is in the right place?), it is set to false when there are other players joined and set to true when there are no RemotePlayerModel.  
So if the host stays on ESC screen, the game will automatically pause when all connections are closed.
- Change GameMain_Transpiler to deal with both `_paused` and `_fullscreenPaused` (toggle by Run game in dyson editor)

### DIFF
--- a/NebulaPatcher/Patches/Transpilers/GameMain_Transpiler.cs
+++ b/NebulaPatcher/Patches/Transpilers/GameMain_Transpiler.cs
@@ -8,11 +8,11 @@ namespace NebulaPatcher.Patches.Transpiler
     [HarmonyPatch(typeof(GameMain))]
     class GameMain_Transpiler
     {
-        //Enable Pausing only when Multiplayer.CanPause is true:
+        //Enable Pausing only when there is no session or  Multiplayer.Session.CanPause is true:
         //Change:  if (!this._paused)
-        //To:      if (!(this._paused && Multiplayer.CanPause))
+        //To:      if (!(this._paused && (Multiplayer.Session == null || Multiplayer.Session.CanPause)))
         //Change:  if (this._fullscreenPaused && !this._fullscreenPausedUnlockOneFrame)
-        //To:      if (this._fullscreenPaused && Multiplayer.CanPause && !this._fullscreenPausedUnlockOneFrame)
+        //To:      if (this._fullscreenPaused && (Multiplayer.Session == null || Multiplayer.Session.CanPause) && !this._fullscreenPausedUnlockOneFrame)
 
         [HarmonyTranspiler]
         [HarmonyPatch(nameof(GameMain.FixedUpdate))]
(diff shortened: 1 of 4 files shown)


PhantomGamers:
This should probably be an option in NebulaModel.MultiplayerOptions

NebulaPatcher/Patches/Transpilers/GameMain_Transpiler.cs
```diff
@@ -8,46 +8,55 @@ namespace NebulaPatcher.Patches.Transpiler
...
+            {
+                NebulaModel.Logger.Log.Error("GameMain.FixedUpdate_Transpiler failed. Mod version not compatible with game version.");
+                return instructions;
+            }
 
-                    //Change jump of first condition
-                    codes[i + 3] = new CodeInstruction(OpCodes.Brfalse_S, targetLabel);
+            codeMatcher
+                .InsertAndAdvance(
+                    new CodeInstruction(OpCodes.Call, AccessTools.Method(typeof(Multiplayer), "get_" + nameof(Multiplayer.CanPause))),
```
PhantomGamers, line 37: This should be `AccessTools.DeclaredPropertyGetter(typeof(Multiplayer), nameof(Multiplayer.CanPause))`

(though again this should probably be an option in the settings menu so -> MultiplayerOptions.PauseWhenNoConnections (idk what it should be called))

hubastard: Actually, I thought the same at first, but I don't think that we need to add this to the option menu since the only way to pause the game will be if you are in the game alone and in the ESC menu, I don't really think that there is a gain here to keep running the game in background if that's the case.

hubastard:
You should move the canPause boolean in the MultiplayerSession instead of in the Multiplayer class, since it is something that should be reset at each session start.

NebulaWorld/Multiplayer.cs
```diff
@@ -59,5 +68,12 @@ public static void LeaveGame()
                 multiplayerMenu.gameObject.SetActive(true);
             }
         }
+
+        private static void SetCanPause(bool status)
+        {
+            canPause = status;
+            //Tell the user if the game is paused or not
+            GameObject.Find("UI Root/Overlay Canvas/In Game/Esc Menu/pause-text").SetActive(status);
```
starfi5h, line 76: Maybe we can provide some information here in the future?

NebulaPatcher/Patches/Transpilers/GameMain_Transpiler.cs
```diff
@@ -8,46 +8,63 @@ namespace NebulaPatcher.Patches.Transpiler
...
+            if (codeMatcher.IsInvalid)
+            {
+                NebulaModel.Logger.Log.Error("GameMain.FixedUpdate_Transpiler failed. Mod version not compatible with game version.");
+                return instructions;
+            }
+            var skipLabel1 = codeMatcher.Instruction.operand;
+            codeMatcher
+                .CreateLabelAt(codeMatcher.Pos + 1, out Label nextLabel1)
+                .InsertAndAdvance(
+                    new CodeInstruction(OpCodes.Brfalse_S, nextLabel1), //_pasue == false => enter loop
```
PhantomGamers, line 38: typo _pasue -> _pause

NebulaPatcher/Patches/Transpilers/GameMain_Transpiler.cs
```diff
@@ -8,46 +8,63 @@ namespace NebulaPatcher.Patches.Transpiler
...
+                    new CodeInstruction(OpCodes.Call, AccessTools.DeclaredPropertyGetter(typeof(Multiplayer), nameof(Multiplayer.Session))),
+                    new CodeInstruction(OpCodes.Brfalse_S, skipLabel1), //_pasue == true && Multiplayer.Session == null => can pause, skip loop
```
PhantomGamers, line 40: typo _pasue -> _pause

NebulaWorld/MultiplayerSession.cs
```diff
@@ -62,6 +72,8 @@ public MultiplayerSession(NetworkProvider networkProvider)
 
         public void Dispose()
         {
+            CanPause = true;
```
PhantomGamers, line 75: this is unneeded as MultiplayerSession is an instance, so after it is disposed the next time it is seen will be a new instance with fields that have default values

PhantomGamers (changes requested):
Other than these minor things this looks good to me

PhantomGamers (approved):

Issue #437, opened 2021-09-10
Planet Terrain not synced on load


PhantomGamers:
Planet terrains are only synced when the client loads a star initially, and then is not resynced when the client first lands on a planet.
![image](https://user-images.githubusercontent.com/844685/132925595-f1540928-91c1-48b8-9011-80913b48a9f4.png)

sp00ktober:
Could it be eough to call this for clients? https://github.com/hubastard/nebula/blob/b636bdf26649a806ea67f33ecaaee875b2eac56f/NebulaPatcher/Patches/Dynamic/PlanetModelingManager_Patch.cs#L94

PhantomGamers:
I tried that but it tended to result in the planet disappearing for me and then a bunch of errors related to the player position (so maybe the planet didn't disappear and the player just got yeeted somewhere else). It was very strange, but I didn't look into it much further

PR #439, opened 2021-09-12
0.5.1

PhantomGamers:
0.5.1:

- Fixed cases where a multiplayer session could hang on the player joining screen.
- Fixed issue where foundations built by clients would not sync to other clients.
- Fixed issue where the user would not be informed if they were kicked due to a mod mismatch.
- Enabled pausing in Multiplayer when no clients are connected. (thanks to @starfi5h)
- Edited description of api for thunderstore. (No changes to the actual API)


PhantomGamers:
I wanted to fix #437 before doing this but I'm currently unable to figure out how to do that, might as well get these fixes out to users in the mean time

PR #450, opened 2021-09-26
Refactor PlayerData serialization

PhantomGamers:
- adds revision information for versioning
- removes data that is set on connection (username, playerid, mechacolors)

This currently doesn't work as intended as when loading playerdata from 0.6.1
> ArgumentOutOfRangeException: Index and count must refer to a location within the buffer.

is thrown by https://github.com/hubastard/nebula/blob/8f3fadf49fc8e0de472f3559fe6e6a3acdca0d1f/NebulaNetwork/SaveManager.cs#L98 and I'm not quite sure why


PhantomGamers:
ignore this I'm dumb

PR #459, opened 2021-10-03
Fix NRE in ACH_BroadcastStar.OnGameTick()

PhantomGamers:
NRE can occur while client is in outer space (i.e. not in a planetary system), so return early if an NRE would have occurred


sp00ktober (approved):
looks good to me.

PR #465, opened 2021-10-27
Valve sockets networking

maximegmd:
This PR replaces websockets with [GameNetworkingSockets](https://github.com/ValveSoftware/GameNetworkingSockets/) by Valve.

A few caveats with this PR:
- There is an issue with GameNetworkingSockets (see https://github.com/ValveSoftware/GameNetworkingSockets/issues/198) that is worked around by polling manually every frame, so until there is a fix for this, polling manually will have to do.
- This requires the manual management of big packet fragmentation, it also involves copying packets around a lot, I am not knowledgeable enough in C# to know if and how this can be optimized.

I benchmarked the library and managed to get 8.5MB/s of throughput (hard limit of GNS, will PR an increase with Valve).

It's supposed to work fine on other platforms, untested.

### DIFF
--- a/.github/scripts/thunderstore_bundle.js
+++ b/.github/scripts/thunderstore_bundle.js
@@ -124,7 +124,7 @@ function generateManifest() {
       `nebula-${apiPluginInfo.name}-${apiPluginInfo.version}`,
       "PhantomGamers-IlLine-1.0.0",
       "CommonAPI-CommonAPI-1.5.0",
-      "starfi5h-BulletTime-1.2.4",
+      "starfi5h-BulletTime-1.2.5",
     ],
     website_url: "https://github.com/hubastard/nebula"
   };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## Changelog
 
+0.8.6:
+
+- @starfish: Bugfix regarding NRE exception in UpdateDirtyMeshes()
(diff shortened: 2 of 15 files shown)


mmjr-x:
I have compiled the latest master (this time through visual studio) with USE_STEAMWEBRTC (ICE Support):
[GameNetworkingSockets-USE_STEAMWEBRTC-win-x64-Release-e4601df.zip](https://github.com/hubastard/nebula/files/7446231/GameNetworkingSockets-USE_STEAMWEBRTC-win-x64-Release-e4601df.zip)

 aesgcmtestvectors, GameNetworkingSockets.dll.manifest and libssl-1_1-x64.dll are problably not required (although I am not sure about libssl-1_1-x64.dll ) however I just included the complete build output. I build it in Release mode, however I can provide one with debug if requested :)

PhantomGamers:
Did some testing with this today and sadly it still seems to be borked with bigger factories. If the client tries to connect to a planet with a large factory it will just hang on requesting factory data, and if a client connects to an empty planet and then travels to a larger planet the planet will never load in (and humorously the player's mecha will be tied to that planet forever, and if they try to go further than 1000 meters away the camera will disconnect from the mecha).

Another note is that the size of the factory that this happens with varies from client to client. I was able to load in to a particular planet that another client was unable to. I didn't have the proper logging enabled to see what the size of the planets were at the time so this isn't super helpful but it's just something to note.

PR #469, opened 2021-11-04
Prevents index out of bounds in ILSShipManager

sp00ktober:
Adjusted an index check such that it cannot fall below 0 anymore.


PhantomGamers (approved):
Worth merging IMO but eventually we need to figure out what the cause of this desync is in the first place and find a better fix

sp00ktober:
Need to find root cause for this but merging now to prevent error spam for those who run into it.

PR #474, opened 2021-11-12
fix error when upgrading blueprint previews

PhantomGamers:


hubastard (approved):

NebulaNetwork/PacketProcessors/Factory/Entity/UpgradeEntityRequestProcessor.cs
```diff
@@ -31,17 +31,32 @@ public override void ProcessPacket(UpgradeEntityRequest packet, NebulaConnection
 
                 Quaternion qRot = new Quaternion(packet.rot.x, packet.rot.y, packet.rot.z, packet.rot.w);
                 Vector3 vPos = new Vector3(packet.pos.x, packet.pos.y, packet.pos.z);
-                EntityData[] ePool = planet.factory.entityPool;
 
-                for (int i = 1; i < planet.factory.entityCursor; i++)
+                
+
+                for (int i = 1; i < (packet.IsPrebuild ? planet.factory.prebuildCursor : planet.factory.entityCursor); i++)
```
Therzok, line 37: prefer setting this before the loop, otherwise you're switching on packet.IsPrebuild on every loop iteration

Therzok: Or better yet, why not have 2 different loops here?

Therzok (commented):
Thanks for the changes!

PR #481, opened 2021-11-24
Quickfix for IndexOutOfBounds due to WarningSystem

sp00ktober:
Avoid index out of bounds by skipping a method. This is a quickfix but needs to be replaced by a proper syncing of the warning system introduced in the latest game patches.

grafik

`id` is not synced.


PhantomGamers:
Wouldn't it be better to skip this entirely for clients until we have sync for it? Also the method that adds warning data

sp00ktober:
As its now players can configure their own warnings and remove them again, but they dont get propagated to other players and for clients they get reset when leaving the star system.
This patch just makes sure that no invalid array entry is accessed by some other calls.

but i guess we could also just skip it entirely if thats preferred :)

hubastard (approved):

PR #484, opened 2021-12-02
Fix for 0.8.23.9989

PhantomGamers:
This is working now. While testing this I have run into an issue where the third client I connect does not see the 2nd client, however I am also able to reproduce this with 0.7.4 and other people have successfully played with 3+ people on that version so I think this is not a blocking issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+0.7.5:
+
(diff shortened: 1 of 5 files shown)


PhantomGamers:
With 0a0ee11 this is the result, need to figure out why the armor itself does not render 🤔
image

sp00ktober (approved):

PR #486, opened 2021-12-05
Hotfix for error if one player is in space

UnrealValentin:
Sometimes if one player is in space, the others will also have localPlanet as null for some reason. 
Needs further investigation..

Feel free to request edits.

Edit: I am referring to this Stacktrace a user reported while one client was in space, other was putting stuff in chests:
image
Neither of them was a host. See the #bug-reports channel in the discord.


sp00ktober:
I just tested this with 2 players (client in space host at planet). Placing something inside the storage did not trigger the error for the client, changing the storage size or placing new storages did not trigger it too.
Im not sure how this error could be triggered as afaik `localPlanet` should always be set when that code is run (players in space can't interact with any storage and thus should always skip that patch)

But im fine with having an additional check in there :)

sp00ktober (approved):

PhantomGamers:
Seems like a really weird issue, whenever a client triggers this packet to be sent it should be guaranteed that the event is generated on their end (since we check against incomingevents), and in this case I don't see how the localplanet can be null... hmmm, still seems fine to add this check

PR #487, opened 2021-12-06
Another container bug

CutestNekoAqua:
Another bug encountered with nebula and containers. The original error was thrown in stuff we don't modify, but we can prevent that function from getting called, when there is no planet around. At least I hope so, as its a hard to reproduce bug, again.
Screenshot: 
image

### DIFF
--- a/NebulaWorld/Logistics/ILSShipManager.cs
+++ b/NebulaWorld/Logistics/ILSShipManager.cs
@@ -268,6 +268,9 @@ namespace NebulaWorld.Logistics
                 UIStationStorage[] stationStorageUI = stationWindow.storageUIs;
                 if (stationStorageUI != null && stationStorageUI.Length > storageIndex)
                 {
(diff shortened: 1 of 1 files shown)


NebulaWorld/Logistics/ILSShipManager.cs
```diff
@@ -268,6 +268,9 @@ private void RefreshValuesUI(StationComponent stationComponent, int storageIndex
                 UIStationStorage[] stationStorageUI = stationWindow.storageUIs;
                 if (stationStorageUI != null && stationStorageUI.Length > storageIndex)
                 {
+                    //Fixes a bug where the game thinks you're still in a container while exiting the planet, hopefully
```
UnrealValentin, line 271: Notice that this bug is not guaranteed to be 100% fixed. xD

starfi5h:
Sorry, it's outdated now due to the rework of ILSShipManager.

PR #489, opened 2021-12-22
Fix ray receiver syncing, player lighting, research refund

starfi5h:
- Add syncing of dyson sphere status so now ray receivers will have the correct output.
- Fix the lighting of remote player armor.
- Fix an issue where client will get duplicate items when canceling current manual research.


PhantomGamers (approved):
Looks good to me, nice work!

sp00ktober:
Looks good to me too, thanks! :)

sp00ktober (approved):

PR #492, opened 2022-01-05
Fix ping indicator + other small changes

starfi5h:
- Fix ping indicator so now it will show the average round trip time.
- Store PlayerAnimationUpdate in buffer to match the timing of movement update.
- Now host only send DysonSphereStatusPacket  if the sphere is not empty, or the system has power consumers(ray receivers).

### DIFF
--- a/NebulaNetwork/Server.cs
+++ b/NebulaNetwork/Server.cs
@@ -167,7 +167,7 @@ namespace NebulaNetwork
                     DysonSphere dysonSphere = dysonSpheres[i];
                     if (dysonSphere != null && (dysonSphere.energyReqCurrentTick + dysonSphere.energyGenCurrentTick > 0))
                     {
-                        SendPacketToStar(new DysonSphereStatusPacket(dysonSpheres[i]), dysonSpheres[i].starData.id);
+                        SendPacketToStar(new DysonSphereStatusPacket(dysonSphere), dysonSphere.starData.id);
                     }
                 }
             }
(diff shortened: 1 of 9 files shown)


NebulaNetwork/Server.cs
```diff
@@ -164,7 +164,8 @@ public override void Update()
                 DysonSphere[] dysonSpheres = GameMain.data.dysonSpheres;
                 for (int i = 0; i < dysonSpheres.Length; i++)
                 {
-                    if (dysonSpheres[i] != null)
+                    DysonSphere dysonSphere = dysonSpheres[i];
+                    if (dysonSphere != null && (dysonSphere.energyReqCurrentTick + dysonSphere.energyGenCurrentTick > 0))
                     {
                         SendPacketToStar(new DysonSphereStatusPacket(dysonSpheres[i]), dysonSpheres[i].starData.id);
```
Therzok, line 170: use dysonSphere as well here.

NebulaWorld/SimulatedWorld.cs
```diff
@@ -205,8 +205,6 @@ public void UpdateRemotePlayerAnimation(PlayerAnimationUpdate packet)
                 if (remotePlayersModels.TryGetValue(packet.PlayerId, out RemotePlayerModel player))
                 {
                     player.Animator.UpdateState(packet);
-
-                    player.Effects.UpdateState(packet);
```
hubastard, line 209: Isn't this needed to play the VFX like mining torch and thrusters?

starfi5h: I move the function call to RemotePlayerAnimation.UpdateState() so they can share the same packet buffer.

hubastard (approved):

PR #496, opened 2022-01-10
Add WarningSystem syncing

starfi5h:
- sync alarm switch on the building (PlanetFactory.EnableEnityWarning)
- sync traffic monitor (MonitorComponent & SpeakerComponent)
- host updates global alarm signals when it's necessary every 1s. (warningSignals, warningCounts)
- client sends requests when it needs detailed information. (warningPool)

~~Known issue: "No connection to the power grid", "No power supply to the power grid" can not show in host when client turn on alarm on a different planet.~~ Found the cause, the game set ``isActive`` to false on the remote planet so PowerSystem.GameTick() can't update signType.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/PowerSystem_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/PowerSystem_Patch.cs
@@ -11,7 +11,10 @@ namespace NebulaPatcher.Patches.Dynamic
         public static void PowerSystem_GameTick_Prefix(long time, ref bool isActive)
         {
             //Enable signType update on remote planet every 64 tick
-            isActive |= (time & 63) == 0;
+            if ((time & 63) == 0 && Multiplayer.IsActive && Multiplayer.Session.LocalPlayer.IsHost)
+            {
+                isActive |= true;
+            }
         }
 
         [HarmonyPostfix]
--- a/NebulaPatcher/Patches/Dynamic/PowerSystem_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/PowerSystem_Patch.cs
@@ -6,6 +6,14 @@ namespace NebulaPatcher.Patches.Dynamic
     [HarmonyPatch(typeof(PowerSystem))]
     internal class PowerSystem_Patch
     {
+        [HarmonyPrefix]
+        [HarmonyPatch(nameof(PowerSystem.GameTick))]
+        public static void PowerSystem_GameTick_Prefix(long time, ref bool isActive)
+        {
+            //Enable signType update on remote planet every 64 tick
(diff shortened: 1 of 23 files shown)


PhantomGamers:
One issue I'm noticing is that under the warning count where it shows the building that is generating the warning, the client does not see the first building that is firing a warning. If only one building is firing a warning the client will just see an empty gray box, and if more are firing the client with see numWarnings - 1.

image

IMO this isn't really a blocking issue and this is worth merging anyway, that can always be fixed later. I tried to find what could be causing this but I'm not even really sure which code is responsible for that part of the UI.

starfi5h:
It was WarningSystem loops warningPool from index-1 but I imported warningPool from index-0, so the first warningData is missing on the client. Will fix in the next commit.

PhantomGamers:
> It was WarningSystem loops warningPool from index-1 but I imported warningPool from index-0, so the first warningData is missing on the client. Get it fixed in this commit.

Ah nice find! Those dang arrays that start at index 1 are tricky.

NebulaPatcher/Patches/Dynamic/PowerSystem_Patch.cs
```diff
@@ -6,6 +6,14 @@ namespace NebulaPatcher.Patches.Dynamic
     [HarmonyPatch(typeof(PowerSystem))]
     internal class PowerSystem_Patch
     {
+        [HarmonyPrefix]
+        [HarmonyPatch(nameof(PowerSystem.GameTick))]
+        public static void PowerSystem_GameTick_Prefix(long time, ref bool isActive)
+        {
+            //Enable signType update on remote planet every 64 tick
+            isActive |= (time & 63) == 0;
+        }
```
PhantomGamers, line 15:
```suggestion
        [HarmonyPrefix]
        [HarmonyPatch(nameof(PowerSystem.GameTick))]
        public static void PowerSystem_GameTick_Prefix(long time, ref bool isActive)
        {
        	if(Multiplayer.IsActive)
        	{
            	//Enable signType update on remote planet every 64 tick
            	isActive |= (time & 63) == 0;
        	}
        }
```

PhantomGamers: In addition to this, is this needed for only the host? Only the client? Both?

PhantomGamers:
Other than that suggestion this looks good to me.

Initially I was concerned that the performance impact of e8b2fe7 might have been worse in a save with a lot of warning systems but looking at it now it's the power system that is affected so it should have a constant impact basically? I didn't notice any impact on my machine.

PhantomGamers (approved):

PR #500, opened 2022-01-15
fix universe exploration setting + 0.7.10

### LINKED ISSUE #497: Upgrade Universe exploration (4) perk doesn't work when the mod is activated
Hello,

In a game with only this mod and its dependencies activated, and the upgrade "Universe exploration (4)", it's impossible to view the resources from all planetary systems.
We have the same issue if we load a save in solo or multiplayer mode.

If we load the same save in vanilla, the perk works as intended.

PhantomGamers:
closes #497

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/UIPlanetDetail_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/UIPlanetDetail_Patch.cs
@@ -9,8 +9,6 @@ namespace NebulaPatcher.Patches.Dynamic
     [HarmonyPatch(typeof(UIPlanetDetail))]
     internal class UIPlanetDetail_Patch
     {
-        private static int BackupUniverseObserveLevel = -1;
-
         [HarmonyPostfix]
         [HarmonyPatch(nameof(UIPlanetDetail.OnNameInputEndEdit))]
         public static void OnNameInputEndEdit_Postfix(UIPlanetDetail __instance)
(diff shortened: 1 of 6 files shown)


PhantomGamers:
closes #497

NebulaPatcher/Patches/Dynamic/UIPlanetDetail_Patch.cs
```diff
@@ -71,11 +68,26 @@ public static bool OnPlanetDataSet_Prefix(UIPlanetDetail __instance)
...
+            {
+                GameMain.history.universeObserveLevel = GetUniverseObserveLevel();
+            }
+        }
+
+        private static int GetUniverseObserveLevel()
+        {
+            int level = 0;
+            // the tech ids of the 4 tiers of Universe Exploration from https://dsp-wiki.com/Upgrades
+            for (int i = 4104; i >= 4101; i--)
```
mattsemar, line 81: Another option would be to use `i = 4` and then do `...TechUnlocked(i + 4100))`, then you could just `return i - 1` without needing mod operator.

PhantomGamers: Yeah but then it has to do addition each time while the mod operator only has to get called once.

Ultimately doesn't make much of a difference idk, I liked doing it this way so that it's clear that it's just going through the tech ids 4101-4104 in the loop itself

sp00ktober (approved):
Seems to work now on the broken save i tested with, and also seems to behave normally on my own save. Thanks for fixing this :D

Issue #230, opened 2021-04-26
ILS wrong remoteOrder value


sp00ktober:
For most of the time the value is correctly displayed (the indication of items that are in transfer). It is also updated to the correct one when the ILS UI is closed and opened again.
But the value is wrong atleast in the following situation:

- ILS on planet A (has ships)
- ILS on planet B (has no ships or none that would pick up work for this)
- ILS on planet A demands items from ILS on planet B
- ILS on planet B indicates correctly how mayn items will be taken out of inventory while ships travel to planet B
- once ships take out the items of ILS on planet B the remoteOrder value does not get updated

sp00ktober:
Im not too sure if this is still an issue as the underlying code did change quite a bit in our latest updates.

starfi5h:
It should be good now? If there are still problems feel free to reopen the issue.

Issue #229, opened 2021-04-26
ILS setting sliders should move when value change


sp00ktober:
The sliders of the settings on the bottom of the ILS UI do not change for other players when they are changed while the other one has the UI open. They get updated after closing and opening the UI.
Values are updated correctly tho, this is just a visual thing.

Issue #210, opened 2021-04-22
GameTick synchronization


Baldie-dev:
Since game tick can go slower on the host (due to the CPU demands), we could maybe improve the GameTick sync on the clients:
If the client's game tick is lower, it would manually trigger extra ones. If the client's game tick is higher, it would skip the next "x" game ticks.

Thoughts?

Therzok:
Yeah, I suggested this since the simulation is pretty deterministic. This would go in well in checksummed and partial resyncs on components.

sp00ktober:
implemented by bullettime

PR #506, opened 2022-01-20
[WIP] Fix compatibility with DSP 0.9

PhantomGamers:
We need to see what broke after the update. here is a quick list of things that come to my mind atm.

- Belts are at least partially broken. picking up items manually does not pick them up for other players, outbound belts from advanced miners do not output stuff for others until they _try_ to build the belts themself, ...
- ILS code has been changed (maybe just a bit?), but its hard to update our transpiler to match the new code. Maybe its easier to rewrite it.
- did not have a look yet at the logic but as the new item enhancement seems to give % chance boosts (afaik) maybe it could be challenging to sync that?
- generally sync the new facilities
- new mecha stuff


NebulaModel/DataStructures/PlayerData.cs
```diff
@@ -45,6 +48,20 @@ public void Serialize(INetDataWriter writer)
...
+            writer.Put(Appearance != null);
+            if(Appearance != null)
+            {
+                using (MemoryStream ms = new MemoryStream())
+                {
+                    using (BinaryWriter wr = new BinaryWriter(ms))
+                    {
+                        Appearance.Export(wr);
+                    }
+                    byte[] export = ms.ToArray();
```
Danil0v3s, line 60: Isn't `Appearance.ToByte()` a shortcut to this?

NebulaModel/DataStructures/PlayerData.cs
```diff
@@ -45,6 +48,20 @@ public void Serialize(INetDataWriter writer)
             Rotation.Serialize(writer);
             BodyRotation.Serialize(writer);
             Mecha.Serialize(writer);
+            writer.Put(Appearance != null);
```
Danil0v3s, line 51: I think you could avoid this line if you use the length to tell whether appearance is there or not

NebulaModel/DataStructures/PlayerData.cs
```diff
@@ -63,6 +80,20 @@ public void Deserialize(INetDataReader reader)
...
+                int len = reader.GetInt();
+                byte[] data = new byte[len];
+                reader.GetBytes(data, len);
+                using (MemoryStream ms = new MemoryStream(data))
+                using (BinaryReader br = new BinaryReader(ms))
+                {
+                    Appearance = new MechaAppearance();
+                    Appearance.Init();
+                    Appearance.Import(br);
+                }
```
Danil0v3s, line 95: Can't you use `Appearance.FromBytes(data)`?

NebulaModel/DataStructures/PlayerData.cs
sp00ktober: It is, could probably be exchanged but would have the same effect.

NebulaModel/DataStructures/PlayerData.cs
sp00ktober: I just took the logic from here: https://github.com/hubastard/nebula/blob/3c7da1bd62a9e58d77eac2f898c07f2fc2d9ccec/NebulaModel/DataStructures/MechaData.cs#L48

NebulaModel/DataStructures/PlayerData.cs
sp00ktober: Again, yes. But its just another way to write the same thing.

sp00ktober:
Im putting a pre release version here and on our discord server now. Maybe we can catch some more bugs before the actual release :D

[nebula_0.8.0_pre-release.zip](https://github.com/hubastard/nebula/files/7997635/nebula_0.8.0_pre-release.zip)

sp00ktober:
closes #230 
closes #229 
closes #210

Jaymus3:
Tested the mod with a couple of people.  So far, everything seems to work properly in the starting system.  Got this error when one player disconnected on the other client that was still connected.  I've also attached the log from that client.  (If this isn't where you'd like pre-release issues reported, feel free to point me to where you'd rather.)

[LogOutput.log](https://github.com/hubastard/nebula/files/8020818/LogOutput.log)

sp00ktober:
Totally fine to report here i guess.
There where some other people reporting this too, i wonder if we should handle undefined packets more gracefully

HoshixD:
I did ~ 10h of testing with one of my friends. 
So far the mod works like a charm. Some feedback:

- We encountered the error of the post above several times (only host). Does not seem to break the game though.
- We encountered 3 instances of the game crashing the Laptop of my friend. We can't say whether that is due to the DSP-update or some changes you guys implemented. 
- There are still some ship artifacts left in orbit (only client) if a player is in transition between systems and the ships use warp cells. Definitely got better though...
- Duping seems to be fixed. We found no way to simulate the environment that would allow for the client to dupe items by manually placing them on a belt leading to a storage unit. This no longer works no matter the latency of the host/client. 

Thanks for the amount of work and time you put into this. Attached is the log file of the host. I can provide the client file if needed.
[LogOutput.log](https://github.com/hubastard/nebula/files/8024006/LogOutput.log)

sp00ktober (approved):

PR #513, opened 2022-01-28
Refactor StationUIManager (0.9.24)

starfi5h:
Refactor StationUIManager
- Remove subscription system. Now update packets are broadcast to every client on that star system.
- Remove input mimic. Now all updates are done in the background and call ``OnStationIdChange()`` to refresh UI.
- Remove ``UIIsSyncedStage``. Disable ``UpdateCooldown`` for now.
- The client initial syncing flow:
[Client] ``UIStationWindow._OnOpen()``: Hide control panel, send ``StationUIInitialSyncRequest`` to Host
[Host]  Get requested ``stationComponent`` data, send ``StationUIInitialSync`` to Client
[Client] Update ``stationComponent``, call ``OnStationIdChange()`` to refresh UI.

Change sliders dragging behaviors

- Make setting sliders only send packets when the left mouse button is released
- Make storage max sliders only apply value when dragging stop
Now sliders should send packets less frequently,

Change itemIcon clicking behaviors

- Change to use original method, and on postfix update the changed item ``count``.
- Client will revert item ``count`` changes and wait for host approval.
- AddOrRemoveItemFromStorage now include item ``inc``
- Drone & Ship count update now try to set the total count of working + idle,  
  if the total count exceeds the limit or not enough idle ships, host will correct the setting value.


sp00ktober:
Looks good, gonna give it a try later when im back at my pc. Thanks for your work!

PR #514, opened 2022-01-29
Add Sync Ups multiplayer option

### LINKED ISSUE #210: GameTick synchronization

starfi5h:
Try to adjust client's ups in the range of 30~120 to match with host when the option is enabled.  
The game tick difference of client and host is calculated by packet round-trip-time, host ``gameTick``, and host ``currentUPS``.  
The difference will be amortized over a period of time by increasing/decreasing client's ups, unless it exceeds the limit.  

Close #210

### DIFF
--- a/NebulaModel/MultiplayerOptions.cs
+++ b/NebulaModel/MultiplayerOptions.cs
@@ -23,6 +23,9 @@ namespace NebulaModel
 
         public string LastIP { get; set; } = string.Empty;
 
+        [DisplayName("Sync Ups")]
+        public bool SyncUps { get; set; } = true;
+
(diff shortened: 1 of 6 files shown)


sp00ktober:
Just tested this merged into the current 0.9 fix branch and it made ils ships stay in sync for a long time (most likely forever but i only tested for 15 minutes).

So seems to work really well :D

PR #520, opened 2022-02-08
Change send and receive behaviors

starfi5h:
**NebulaConnection**

Switch to ``SendAsync`` and add sending packet queuing.

**NetPacketProcessor**

Add ``enable`` property, can pause or resume packet processor.  
It is used to pause packet processing when loading factory data.  

**PlayerManager**

Syncing player now will be added to ``connectedPlayers`` at the moment of requesting factory data, so he can receive the following factory update.

### DIFF
--- a/NebulaModel/Networking/NebulaConnection.cs
+++ b/NebulaModel/Networking/NebulaConnection.cs
@@ -2,6 +2,7 @@
 using NebulaModel.Logger;
 using NebulaModel.Networking.Serialization;
 using System;
+using System.Collections.Generic;
 using System.Net;
 using WebSocketSharp;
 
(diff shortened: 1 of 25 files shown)


NebulaModel/Networking/NebulaConnection.cs
```diff
@@ -24,25 +27,47 @@ public NebulaConnection(WebSocket peerSocket, EndPoint peerEndpoint, NetPacketPr
...
+            }            
+        }
+
+        public void ProcessPacketQueue()
+        {
+            if (pendingPackets.Count == 0 || !enable)
+            {
+                return;
+            }
+            byte[] packet = pendingPackets.Dequeue();
```
Therzok, line 53: This needs to be run under the lock.

Either make the method private or take while using the collection.

Alternatively, you can use a ConcurrentQueue.

starfi5h: But I've already enclosed all 3 positions calling ``ProcessPacketQueue()`` in the lock.  Should I find another way to write these functions?  
What I hope to achieve is to minimum locks use and also preserve the packet oreder.

Therzok: Yeah, but the API is public is what I was getting to. Anything can end up calling this, and it wouldn't necessarily have the lock taken. You could just make the method private.

sp00ktober (approved):
Worked really good with the BulletTime.dll supplied in our discord server!
This is really cool :D

PR #521, opened 2022-02-09
Bugfix and issue list for 0.8.0 pre-release

starfi5h:
A list of issues solved/unsolved in 0.8.0 pre-release version

Resolved, will be fixed in the next update:
-------------

- Sprayer and monitor are not connected to belt when client builds on a different planet.

> (Self note: watch out ``OnBeltBuilt()`` and ``OnAddonBuilt()``, they use ``planet.physics.nearColliderLogic`` which is not synced, so it is needed to sync the result of these functions)

- On client, sprayer and monitor will build to the wrong place when there are multiple prebuilds.

- Other players get items refunds too when a player changes the recipe of a lab with matrix in it.

- Other clients will receive an undefined packet when a client leaves.

- When a player disconnects during loading, the host will get NRE or the whole game pause.

- If a loaded remote planet has ILS built by other players, the ILS will not sync on this player and may cause errors when they change its storage.

- Player color can show differently on other players ([fix in the main branch](https://github.com/hubastard/nebula/commit/40b2408aa2c9eada7f88978ea5cecb84d8564aad) by sp00ktober)

- Clients lose their inventory when loading an old .server file
=> Now support backward compatibility to revision 4 (current: 5)

- When a client opens dyson editor without any dyson spheres loaded, it will get an error.
=> Now dyson spheres only unload when arriving a new system

- When clients connect sorters output by placing belts under the sorters, the sorters will stop working after host reload the game.  

Known issues:
-------------

- When client land on a heavily built-upon planet, it may trigger NRE on renderer
> NRE: CargoTraffic.AlterBeltRenderer (IL_0350)

Unknown cause, can't reproduce yet:
-------------
- Sometimes when host deletes an ILS on a different planet in the same system, client will get an error.
unknown

- Belts connect randomly on client, fix after reconnection.
DSPGAME_2022-02-07_18-41-30

- Sometimes sorters will stop working. The animation is still running but no cargo is moved.  Fix after replacement.
- Sometimes the splitter filter is randomly set to the wrong item and can't be changed. Have to replace the splitter.
- Sometimes clients can't see the profiler functioning in factories. They don't see the blue bonus bar.

Other changes:
-------------
copy
Add a button to copy and close the error message. Special thanks to Therzok's WhatTheBreak and crecheng's CloseError

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/DESelection_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/DESelection_Patch.cs
@@ -30,35 +30,37 @@ namespace NebulaPatcher.Patches.Dynamic
                                 return true;
                             }
                         }
+                        //No dyson sphere exist, close the UI
+                        UIRoot.instance.uiGame.dysonEditor._Close();
+                        return false;
+                    }
+                    else if (GameMain.data.dysonSpheres[starData.index] == null)
+                    {
+                        //Local dyson sphere hasn't loaded yet, close the UI
+                        UIRoot.instance.uiGame.dysonEditor._Close();
+                        return false;
                     }
                 }
                 if (starData != null && GameMain.data.dysonSpheres[starData.index] == null)
                 {
+                    if (Multiplayer.Session.DysonSpheres.RequestingIndex != -1)
+                    {
+                        InGamePopup.ShowInfo("Loading", $"Loading Dyson sphere {starData.displayName}, please wait...", "OK");
+                        return false;
+                    }
+                    Multiplayer.Session.DysonSpheres.RequestingIndex = starData.index;
                     Log.Info($"Requesting DysonSphere for system {starData.displayName} (Index: {starData.index})");
                     Multiplayer.Session.Network.SendPacket(new DysonSphereLoadRequest(starData.index, DysonSphereRequestEvent.Load));
-                    //Set viewDysonSphere to null until requested dyson sphere data is arrived.
                     __instance.ClearAllSelection();
-                    __instance.viewStar = starData;
-                    __instance.viewDysonSphere = null;
+                    InGamePopup.ShowInfo("Loading", $"Loading Dyson sphere {starData.displayName}, please wait...", "OK");
+                    // Restore comboBox back to original star
+                    UIComboBox dysonBox = UIRoot.instance.uiGame.dysonEditor.controlPanel.topFunction.dysonBox;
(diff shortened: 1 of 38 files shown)


NebulaPatcher/Patches/Dynamic/DESelection_Patch.cs
```diff
@@ -30,35 +30,37 @@ public static bool SetViewStar_Prefix(DESelection __instance, ref StarData starD
...
+                        //Local dyson sphere hasn't loaded yet, close the UI
+                        UIRoot.instance.uiGame.dysonEditor._Close();
+                        return false;
                     }
                 }
                 if (starData != null && GameMain.data.dysonSpheres[starData.index] == null)
                 {
+                    if (Multiplayer.Session.DysonSpheres.RequestingIndex != -1)
+                    {
+                        InGamePopup.ShowInfo("Loading", $"Loading Dyson sphere {starData.displayName}, please wait...", "OK");
```
sp00ktober, line 48: As you call fadeout later, why add the Ok button here? :)

NebulaPatcher/Patches/Dynamic/DESelection_Patch.cs
```diff
@@ -30,35 +30,37 @@ public static bool SetViewStar_Prefix(DESelection __instance, ref StarData starD
...
+                        return false;
+                    }
+                    Multiplayer.Session.DysonSpheres.RequestingIndex = starData.index;
                     Log.Info($"Requesting DysonSphere for system {starData.displayName} (Index: {starData.index})");
                     Multiplayer.Session.Network.SendPacket(new DysonSphereLoadRequest(starData.index, DysonSphereRequestEvent.Load));
-                    //Set viewDysonSphere to null until requested dyson sphere data is arrived.
                     __instance.ClearAllSelection();
-                    __instance.viewStar = starData;
-                    __instance.viewDysonSphere = null;
+                    InGamePopup.ShowInfo("Loading", $"Loading Dyson sphere {starData.displayName}, please wait...", "OK");
```
sp00ktober, line 55: same as above

NebulaPatcher/Patches/Dynamic/DESelection_Patch.cs
starfi5h: some users may be impatient about loading time, so I add a way to let them close the info window. (although Esc can do the same thing)

starfi5h:
OK, this is the last commit in this PR. I don't think entityId will desync if the factory data is corrected transmitted in the first place, but just add in-case so we can rule out entityId desync in future debugging.

PR #529, opened 2022-02-20
Bugfix, Tooltips and minimap indicator

sp00ktober:
So far contains:

- Added Tooltips to the Nebula ingame settings
- Added fix and setting for error `System.ObjectDisposedException` on host resulting in random client disconnect
- Added code to handle rare IndexOutOfBounds in `PlanetFactory.Import()`
- Fix wrong array size in ILS storage and slot
- Added remote player minimap indicator
- Add command to navigate to planet, star or player
- Add command to list planets in a solar system
- Listing players now also displays their id

grafik
grafik
grafik


PhantomGamers (changes requested):
If you trigger a tooltip to pop up and then close the menu, the tooltip persists on the screen indefinitely 
image

Also I think the /system command should use the system the player is currently in (if they are in one) if it is used without any arguments.

Other than those two things this looks good!

PhantomGamers (approved):
LGTM

NebulaWorld/Chat/Commands/NavigateCommandHandler.cs
```diff
@@ -88,7 +100,7 @@ public string GetDescription()
         }
         public string GetUsage()
         {
-            return "[planet name | planet id | star name | star id | clear | player] [player id | playr name]";
+            return "[planet name | planet id | star name | star id | clear]\r\nUsage: /n player [player id | playr name]";
```
kremnev8, line 103: I don’t think this solution is good. Instead change this method to return an array and change help command.

kremnev8 (changes requested):

NebulaWorld/Chat/Commands/HelpCommandHandler.cs
```diff
@@ -43,9 +43,16 @@ private static string GetCommandDetailsOutput(string commandName)
             string aliasList = ChatCommandRegistry.GetCommandAliases(commandName).Join(FullCommandName);
 
             StringBuilder sb = new StringBuilder();
+            foreach(string usage in handler.GetUsage())
+            {
+                sb.Append($"Usage: {FullCommandName(commandName)} {usage}\n");
+            }
+            string usageString = sb.ToString();
```
kremnev8, line 50: This is a weird way of doing this. Move iterations down and save one string variable allocated.

kremnev8 (changes requested):

Issue #541, opened 2022-04-24
ILS/PLS desync issues


sp00ktober:
It seems that there are still some cases left where it is possible to:

- loose or at least not see drones/vessels as client when putting them into the station
- take out infinte amounts of drones/vessels from stations as client
- loose content sync of stations as client until their ui is opened while staying on the same planet for a while

The latter one can cause belt content desync and then in turn all other sorts of desync issues, so it needs some investigation.

AlienXAXS:
Instead of making a new issue, i feel this report is relevant:

I have only just recently got into Space Warpers with my logistics system, I have noticed quite a lot of ships that don't seem to animate quite right when they should have consumed a space warper (or perhaps they are not, and that's the issue?).

As you can see by the following image, logistics ships that should be warping are not, however their cargo is still being delivered to the other solar system.

![image](https://user-images.githubusercontent.com/1773445/197511131-3261ff64-badf-4012-a53d-4d16c7a89015.png)

Their slow-flying animations will progress until their warper animation would have arrived at the planet that resides in their destination solar system, at which point the ship just disappears and the ILS no longer shows product in transit.

PR #542, opened 2022-04-24
Fix UpdateDirtyMeshes error, ILS desync

starfi5h:
Fix bugs introduced in PR#536 
- fix UpdateDirtyMeshes error when client imports factory
- fix ILS desync when client imports factory (#541)


sp00ktober (approved):
Looks good to me, we will see if it actually fixes he issues for those who reported it.

PR #549, opened 2022-05-01
Upnp pmp support

mmjr-x:
Added Upnp/pmp support, however it probably needs some testing. Hopefully this will help people to not have to do port forwarding manually

For anyone that wants to try, I zipped up the nebula-NebulaMultiplayerMod and nebula-NebulaMultiplayerModApi folders from my BepInEx\plugins here:
[upnp-pmp-support 72d1ade8905b13d51ad132a189cd7693ec40f7cb.zip](https://github.com/hubastard/nebula/files/8599779/upnp-pmp-support.72d1ade8905b13d51ad132a189cd7693ec40f7cb.zip)
.


NebulaModel/MultiplayerOptions.cs
```diff
@@ -16,6 +16,10 @@ public class MultiplayerOptions : ICloneable
         [UIRange(1, ushort.MaxValue)]
         public ushort HostPort { get; set; } = 8469;
 
+        [DisplayName("Enable UPnp/Pmp Support")]
+        [Description("If enabled, attempt to automatically create a port mapping using UPnp/Pmp (only works if your router has this feature and it is enabled)")]
+        public bool EnableUPnpOrPmpSupport { get; set; } = true;
```
starfi5h, line 21: The default option should probably be set to false. Let the user enable this option by themself.

NebulaNetwork/Server.cs
```diff
@@ -66,6 +69,27 @@ public override void Start()
             PacketProcessor.SimulateLatency = true;
 #endif
 
+            if (Config.Options.EnableUPnpOrPmpSupport)
+            {
+                Task.Run(async () => {
+                    var discoverer = new NatDiscoverer();
+                    try
+                    {
+                        var device = await discoverer.DiscoverDeviceAsync();
```
starfi5h, line 77: Hmm... Do you think it is a good idea to put the discover device part somewhere else (e.g. when the option is enabled or when preload completes) so it would take less time to start up server?

starfi5h: I don't have the device so I don't know how long the time it takes. If it normally is really fast then you can ignore this suggestion.

NebulaModel/MultiplayerOptions.cs
mmjr-x: I changed it to false by default

starfi5h (approved):

Issue #548, opened 2022-05-01
非主机玩家进入气巨时会报错


ighlES:
```ini
An error has occurred! Game version 0.9.25.12077 (Client)
Mods use: [EnableMilkyWayGalaxy1.1.2] [Auxilaryfunction1.5.4] [Place the pump anywhere. Pumps whatever ocene type the planet has.1.0.0] [LongSorter1.1.0] [IlLine1.0.0] [LDBTool2.0.5] [DSP Mod Save1.1.4] [DSP Common API1.5.1.0] [NebulaMultiplayerMod0.8.6.2] [NebulaMultiplayerModApi1.2.0.5] [BulletTime1.2.5] [Common API Nebula Compatibility1.5.1.0] [StarMapTools6.4] 
[Nebula Error] System.NullReferenceException: Object reference not set to an instance of an object

  at PlanetData.UpdateDirtyMesh (System.Int32 dirtyIdx) [0x0009e] in <192758684e944be3b703e707f5f46edf>:IL_009E 

  at PlanetData.UpdateDirtyMeshes () [0x00006] in <192758684e944be3b703e707f5f46edf>:IL_0006 

  at NebulaPatcher.Patches.Dynamic.GameData_Patch.OnActivePlanetFactoryLoaded_Prefix (GameData __instance, PlanetData planet) [0x000da] in <2277c262de38488f97bea6bb5aa06bcb>:IL_00DA 


```
![image](https://user-images.githubusercontent.com/61189870/166156957-97f30cae-a24a-4f43-8cc8-3a14b7663614.png)

Issue #546, opened 2022-04-29
Planet type mismatch when hosting an old save


starfi5h:
When hosting an old save, some explored planets type might be different on client side.  
For example, [this save](https://drive.google.com/file/d/1UD9Kb936AnwbacaF7tJhKjlfYThydqTs/view?usp=sharing) was created in game version 0.7.x, In host OsPegasi II is a Waterworld, but in client it is a Sakura Ocean.  
![planet type](https://user-images.githubusercontent.com/50672801/166081643-f2bf654a-132c-4472-ae8c-9067ff75d059.jpg)

PR #551, opened 2022-05-03
Bugfixes for login and loading process

starfi5h:
Bugfixes:
- Fix planet type mismatch (#546)
- Fix error when client loading a gas gaint (#548)
- Fix error when clicking the back button twice in GS2 lobby

Refactor:  
- Because postfix of `MilestoneSystem.SetForNewGame` didn't get called on some machines, I merge `GameHistoryDataRequest`, `TrashSystemRequestDataPacket`, `MilestoneDataRequest` into one request `GlobalGameDataRequest` and call it on `SetForNewGame_Postfix`
- Rename `PingPacket` to `GameStateRequest` and move namespace to `NebulaModel.Packets.GameStates`
- When host arriving on another planet, in order to prevent building packets get processed before the local factory model is loaded, stop packet processor until loading is completed. Hope this change can prevent the following issues: 

host_error_1


PhantomGamers (approved):
Looks good to me, great work!

If you want to push a new version with this make sure to bump the version.json and changelog.md

PR #559, opened 2022-05-13
Small bugfix for UI

starfi5h:
- **Fix a bug that chat window duplicated when player reconnect**  
Make so that chat window persists when players exit multiplayer game.  
   
- **Let AutoPauseEnabled control non-dedicated host pause state**
  
- **Suppress abnormality reports on client**  
Because multiplayer is played on host, the client won't get achievement anyway, so there is no need to inform the player about abnormality triggers.

### DIFF
--- a/NebulaNetwork/Ngrok/NgrokManager.cs
+++ b/NebulaNetwork/Ngrok/NgrokManager.cs
@@ -68,7 +68,7 @@ namespace NebulaNetwork.Ngrok
                     {
                         InGamePopup.ShowWarning(
                             "Ngrok download and installation confirmation",
-                            "Ngrok is support is enabled, however it has not been downloaded and installed yet, do you want to automattically download and install Ngrok?",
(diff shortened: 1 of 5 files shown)


NebulaNetwork/Ngrok/NgrokManager.cs
```diff
@@ -68,7 +68,7 @@ public NgrokManager(int port, string authtoken = null, string region = null)
                     {
                         InGamePopup.ShowWarning(
                             "Ngrok download and installation confirmation",
-                            "Ngrok is support is enabled, however it has not been downloaded and installed yet, do you want to automattically download and install Ngrok?",
+                            "Ngrok support is enabled, however it has not been downloaded and installed yet, do you want to automattically download and install Ngrok?",
```
PhantomGamers, line 71:
```suggestion
                            "Ngrok support is enabled, however it has not been downloaded and installed yet, do you want to automatically download and install Ngrok?",
```

PhantomGamers (changes requested):

PhantomGamers (approved):

PR #553, opened 2022-05-07
Headless mode/dedicated server

starfi5h:
Continue the work of [Baldie-dev](https://github.com/Baldie-dev) from PR #62 

> Possibility to run the server in headless mode without the usage of GPU or any rendering.
> The dedicated server can be started from the command line by:
> `./DSPGAME.exe -batchmode -nographics -server -load "asd"`
> Where "asd" is the name of the savefile.
> In order to run the game from the exe you must create a `steam_appid.txt` file in the root of the game directory (next to DSPGame.exe) with the text `1366540`

**New command parameters**
`-server` : Start DSP in headless server mode. 
`-load <savename>` : Currently headless mode needs to specify the save to load.  
`-ups <integer value>`: Set UPS to value using `FPSController.SetFixUPS(value)`  

**New config options**  
These options are not shown in settings, have to open the config file to edit.  
`AutoPauseEnabled`: Auto pause the game when there is no player connecting. Default: `true`  
`EnableRemoteSaveAccess`: Enable following chat commands. This should only be enabled when all clients are trusted users.   Default: `false`  

**New chat commands**  
Performance server load/save by client commands.  
Only works when server is dedicated and has `EnableRemoteSaveAccess` enabled.  

- `/server list [saveNum]`
sever-list
Show the recent [saveNum] savefiles by time and savename on the server.   

- `/server save [saveName]`
Save current game as [saveName]. If saveName is not specify, it will save to `_lastexit_.dsv`.  

- `/server load <saveName>`
Load game from `saveName`. Current game will be save to `_lastexit_.dsv` before restart.  

----

From observing `DysonSwarm.GameTick()`, dyson sails position and velocity, which are processed by GPU device, are only affecting visual, so we can safely skip updating them. The sails absorbing and expiring are determined by time and processed by CPU.   
  
The downside is dyson swarm might look weird on client, I'm planning to make another mod to compress and reconstruct dyson swarm by sail lifetime histogram and orbit parameters to solve this issue.

PR #563, opened 2022-05-17
Headless mode improvement

starfi5h:
Continue #553 

**Config options**
Add `RemoteAccessPassword`
Change `EnableRemoteSaveAccess` to `RemoteAccessEnabled`

**Chat commands**
`/server login <password>` : if server's `RemoteAccessPassword` is not empty or whitespace, it will require the user to put in password before doing remote access actions. CD is 2 seconds.  
login
`/server info [full]`: this will return server's info to the user.

**Bugfix**
Fix a bug that autosave doesn't work on headless server.  
Make the mecha model of headless server doesn't show up on clients' world.   

**Others**
Log Ngrok address and config options during the game startup.  
Chat now support 10 command history, switch by up and down arrow key.

### DIFF
--- a/NebulaWorld/MonoBehaviours/Local/Chat/ChatWindow.cs
+++ b/NebulaWorld/MonoBehaviours/Local/Chat/ChatWindow.cs
@@ -26,9 +26,11 @@ namespace NebulaWorld.MonoBehaviours.Local
         
         
         internal UIWindowDrag DragTrigger;
-        private Queue<QueuedMessage> outgoingMessages = new Queue<QueuedMessage>(5);
+        private readonly Queue<QueuedMessage> outgoingMessages = new Queue<QueuedMessage>(5);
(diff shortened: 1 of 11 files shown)


NebulaWorld/MonoBehaviours/Local/Chat/ChatWindow.cs
```diff
@@ -42,6 +44,18 @@ void Update()
 
             notifierMask.sizeDelta = new Vector2(chatPanel.rect.width, notifierMask.sizeDelta.y);
 
+            if (chatBox.isFocused)
+            {
+                if (Input.GetKeyDown(KeyCode.UpArrow))
+                {
+                    UseHistoryInput(-1);
```
mattsemar, line 51: Nice

PhantomGamers (approved):

PR #574, opened 2022-06-20
Cleanup + bugfix

starfi5h:
- Organize storage packets. Move them under `Packets.Factory.Storage
- Fix a bug that makes advance miner power usage abnormal.
- Fix error when using reform/revert whole planet in single player.
- Fix an issue that client lost items in manual research when logging out. 
Now when a client logs out, it will try to take back items in mecha lab, and send the current mecha inventory and data to host. It will wait 0.1 seconds to send the packet then disconnect.


PhantomGamers (approved):
LGTM

PR #575, opened 2022-06-20
Enhance user experience

starfi5h:
- Show red error for unhandled exceptions during initial patching, so normal user will know nebula is incompatible.
- Add new chat option NotificationDuration to let user decide how long should chat notification stay on screen.
- Add FragmentInfo to show the download progress percentage
For large packet (factory data, dyson sphere data) during login or arriving another planet

login
arriving planet

Other:
- Preserve NebulaConnection so its state won't be different in PacketProcessor.
- Let MaxUPS and MinUPS can change by other plugins (e.g. BulletTime)


PhantomGamers (approved):
LGTM

PR #577, opened 2022-07-09
Bugfix for dedicated sever sails, and other small changes

starfi5h:
Bugfix
- Fix error when sail capacity increases in dedicated server
- Save player.key to .\Dyson Sphere Program folder when it cannot save to document folder

Others:
- Temporary fix for UpdateDirtyMesh
Now it will skip updating colliders when the mesh is null. Need to investigate why mesh is null in the future
- Add try-catch for NebulaAPI hooked event so it will not break processing flow if it error.

Unsolved Issue:
- Dyson sails are not normal in sandbox mode on dedicated server.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/PlanetData_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/PlanetData_Patch.cs
@@ -20,6 +20,21 @@ namespace NebulaPatcher.Patches.Dynamic
         }
 
         [HarmonyPrefix]
+        [HarmonyPatch(nameof(PlanetData.UpdateDirtyMesh))]
+        public static bool UpdateDirtyMesh_Prefix(PlanetData __instance, int dirtyIdx, ref bool __result)
+        {
+            // Temporary fix: skip function when the mesh is null
+            if (__instance.dirtyFlags[dirtyIdx] && __instance.meshes[dirtyIdx] == null)
+            {
+                Log.Warn(__instance == GameMain.localPlanet ? "Local" : "Remote" + $" PlanetData.UpdateDirtyMesh: meshes[{dirtyIdx}] is null");
+                __result = false;
+                return false;
+            }
+
+            return true;
+        }
+
+        [HarmonyPrefix]
         [HarmonyPatch(nameof(PlanetData.UnloadMeshes))]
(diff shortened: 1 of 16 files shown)


NebulaPatcher/Patches/Dynamic/PlanetData_Patch.cs
```diff
@@ -19,6 +19,21 @@ public static void LoadFactory_Prefix()
             }
         }
 
+        [HarmonyPrefix]
+        [HarmonyPatch(nameof(PlanetData.UpdateDirtyMesh))]
+        public static bool UpdateDirtyMesh_Prefix(PlanetData __instance, int dirtyIdx, ref bool __result)
+        {
+            // Temporary fix: skip function when the mesh is null
```
sp00ktober, line 26: Was there an issue reported where this case was the cause?

starfi5h: In Discord #bug-reports channel:
```
An error has occurred! Game version 0.9.26.13034 (Host)
Mods used: [DSP Mod Save1.1.4] [DSPMarker0.0.6] [IlLine1.0.0] [LDBTool2.0.6] [DSP Common API1.5.4.0] [NebulaMultiplayerMod0.8.10.2] [NebulaMultiplayerModApi1.3.0.7] [BulletTime1.2.6] [Common API Nebula Compatibility1.5.4.0] 
NullReferenceException: Object reference not set to an instance of an object
PlanetData.UpdateDirtyMesh (System.Int32 dirtyIdx);(IL_009E)
PlanetData.UpdateDirtyMeshes ();(IL_0006)
PlanetFactory.FlattenTerrain (UnityEngine.Vector3 pos, UnityEngine.Quaternion rot, UnityEngine.Bounds bound, System.Single fade0, System.Single fade1, System.Boolean removeVein, System.Boolean lift, System.Boolean autoRefresh);(IL_066E)
PlanetFactory.BuildFinally (Player player, System.Int32 prebuildId, System.Boolean autoRefresh);(IL_00A6)
NebulaNetwork.PacketProcessors.Factory.Entity.BuildEntityRequestProcessor.ProcessPacket (NebulaModel.Packets.Factory.BuildEntityRequest packet, NebulaModel.Networking.NebulaConnection conn);(IL_0185)
```

PR #579, opened 2022-09-17
Optimizations + SaveAsLastExit for dedicated server

starfi5h:
- Merge PlayerAnimationUpdate, PlayerUseWarper packet to PlayerMovement packet.
Reduce mecha real-time state update packets from 25/s to 10/s.
- Optimize DroneManager to reduce lag when finding targets.
- Set storage from null to zero-length array for fake ILS.
- Enable SaveAsLastExit for dedicated server
Host can exit gracefully using ctrl-c or crtl-break event


NebulaModel/Utils/NativeInterop.cs
```diff
@@ -17,5 +19,50 @@ public static void HideWindow()
         {
             ShowWindow(GetActiveWindow(), SW_HIDE);
         }
+
+
+        [DllImport("Kernel32")]
+        private static extern bool SetConsoleCtrlHandler(EventHandler handler, bool add);
+
+        private delegate bool EventHandler(CtrlType sig);
```
Therzok, line 27: EventHandler is a very misleading name for this delegate, there's an EventHandler delegate in the BCL.

Maybe name it CtrlHandler or something.

NebulaModel/Utils/NativeInterop.cs
```diff
@@ -17,5 +19,50 @@ public static void HideWindow()
...
+                    Thread.Sleep(1000);
+                    return false;
+            }
+            Thread.Sleep(500);
+            return false;
+        }
+
+        public static void SetConsoleCtrlHandler()
+        {
+            bool result = SetConsoleCtrlHandler(new EventHandler(Handler), true);
```
Therzok, line 64: Even though it's a static method that's being used as the callback, maybe a comment can be added to signal that if the handler is no longer static, it'll can get GC'd because nothing is keeping a reference to the delegate.

NebulaWorld/MonoBehaviours/Remote/RemotePlayerAnimation.cs
```diff
@@ -36,7 +36,7 @@ public void UpdateState(PlayerAnimationUpdate packet)
             PlayerAnimator.jumpWeight = packet.JumpWeight;
             PlayerAnimator.jumpNormalizedTime = packet.JumpNormalizedTime;
             PlayerAnimator.idleAnimIndex = packet.IdleAnimIndex;
-            PlayerAnimator.sailAnimIndex = packet.SailAnimIndex;
+            PlayerAnimator.sailAnimIndex = 0;
```
Therzok, line 39: Won't this stop animating flying remote clients?

NebulaWorld/Player/DroneManager.cs
```diff
@@ -115,18 +106,28 @@ public bool AmIClosestPlayer(ref Vector3 entityPos)
...
-                    if (myDistance > (CachedPositions[model.PlayerId] - entityPos).sqrMagnitude)
-                    {
-                        return false;
-                    }
+                    CachedPositions.Add(model.Movement.GetLastPosition().LocalPlanetPosition.ToVector3());
                 }
             }
+        }
 
+        public bool AmIClosestPlayer(ref Vector3 entityPos)
```
Therzok, line 114: IsLocalPlayerClosestTo()?

NebulaWorld/MonoBehaviours/Remote/RemotePlayerAnimation.cs
starfi5h: SailAnimIndex is used when mecha is flying intersteallar. 0 = normal, 3 = wobbling effect of wrapping, 1 & 2 is related to a CosmicMessage.
I'm not sure if there is a way for a player to see another player closely during sailing in the current game, so I disable it. Will add the syncing back if it is possible.

Therzok: Thanks for the clarification!

Therzok (approved):

PhantomGamers (approved):

PR #612, opened 2023-12-14
Remove CommonAPI dependency

starfi5h:
- Replace CommonAPI functions so no need to install CommonAPI in the future.
- Add `CreateHotkeyControl` to support [KeyboardShortcut](https://docs.bepinex.dev/api/BepInEx.Configuration.KeyboardShortcut.html) type in MultiplayerOptions. 
The available Keycode string can be found in [Unity Doc](https://docs.unity3d.com/2018.4/Documentation/ScriptReference/KeyCode.html). The input text will turn red if it is invalid.  
- Add ChatHotkey, ChatWindowOpacity config options

chat config

### DIFF
--- a/NebulaModel/MultiplayerOptions.cs
+++ b/NebulaModel/MultiplayerOptions.cs
@@ -1,4 +1,5 @@
-﻿using NebulaAPI;
+﻿using BepInEx.Configuration;
+using NebulaAPI;
 using NebulaModel.Attributes;
 using NebulaModel.DataStructures;
 using NebulaModel.Utils;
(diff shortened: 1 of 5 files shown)


PhantomGamers:
Why is CommonAPI still included as a soft dependency then? Do we leverage it at all after this PR?

starfi5h:
The soft dependency is added to preserve the load order that Nebula is loaded after CommonAPI. I'm not sure if CommonAPI has code that is sensitive to load order, so maybe it can be removed?

As for usage, here are some warnings from chatwindow after disable CommonAPI
```
[Warning: Unity Log] The referenced script (CommonAPI.UIWindowResize) on this Behaviour is missing!
[Warning: Unity Log] The referenced script (CommonAPI.MaterialFixer) on this Behaviour is missing!
[Warning: Unity Log] The referenced script on this Behaviour (Game Object 'ChatV2') is missing!
```
- the resize window feature has to use CommonAPI to enable. (Or maybe add the component in this PR?)
- MaterialFixer is to fix TranslucentImage, so the component is removed in the PR.
- I'm not sure the effect of the last one. Haven't tested it out.


So after this PR, Nebula can run without CommonAPI but with less features. Whether to include CommonAPI in thunderstore dependencies can be discussed.

PhantomGamers (approved):
This seems fine to me for now, at least to make developing the update easier on us, but before release we should update the chat assetbundle to remove the dependency on CommonAPI then as well.

PR #615, opened 2023-12-14
Major cleanup

PhantomGamers:
Major changes in this PR.

- Updated .editorconfig to latest config from roslyn repo https://github.com/dotnet/roslyn/blob/fd3194f093806a6b2577a45aa89b0fd60658eee2/.editorconfig
- Reformatted sln with IDE
- Reworked much of the code base to use modern C# features
- Removed a lot of code nesting by inverting if statements
- Swapped target framework on the nebula mod api to netstandard2.0 so that it is compatible with plugins that target either net standard or net framework
- Fixed many namespaces so that they match the directory that they are in

This PR will of course break compatibility with any plugin that is using the API, but since tomorrow's update is bound to break most plugins anyway, and I'm sure we'll have to do more breaking changes to the API, I think this is a good time to do this.


PhantomGamers:
This PR causes this issue
image

I believe it is due to the breaking changes in the API and that the chat asset bundle needs to be rebuilt to target the updated API

starfi5h:
Some issues regarding the chat:

1. Notify messages when login/logoff create an error
```
ArgumentOutOfRangeException: Exception of type 'System.ArgumentOutOfRangeException' was thrown.
Parameter name: messageType
Actual value was SystemInfoMessage.
NebulaWorld.MonoBehaviours.Local.Chat.ChatWindow.SendLocalChatMessage (System.String text, NebulaModel.DataStructures.Chat.ChatMessageType messageType);(0)
NebulaWorld.MonoBehaviours.Local.Chat.ChatManager.SendChatMessage (System.String text, NebulaModel.DataStructures.Chat.ChatMessageType messageType);(0)
NebulaWorld.SimulatedWorld.SendChatMessage (System.String text, NebulaModel.DataStructures.Chat.ChatMessageType messageType);(0)
NebulaWorld.SimulatedWorld.SpawnRemotePlayerModel (NebulaAPI.GameState.IPlayerData playerData);(0)
NebulaWorld.SimulatedWorld.OnPlayerJoinedGame (NebulaAPI.GameState.INebulaPlayer player);(0)
NebulaNetwork.PacketProcessors.Session.SyncCompleteProcessor.ProcessPacket (NebulaModel.Packets.Session.SyncComplete packet, NebulaModel.Networking.NebulaConnection conn);(0)
```
2. Notify now has a dark background
image

starfi5h (approved):

starfi5h (approved):

PR #622, opened 2023-12-21
Add sync for storage filters

### LINKED ISSUE #620: Sync the new StorageComponent.SetFilter(int,int)
`StorageComponent.SetFilter(int,int)` share similarity to `StorageComponent.SetBans(int)`
As how `StorageSyncSetBansPacket` did, create a packet `StorageSyncSetFilterPacket` and sync with a similar method.
If possible, test if operations on storage boxes are all handled and in sync.

Reousa:
Adds synchronization for the new storage filters.

Resolves #620

### DIFF
--- a/NebulaModel/Packets/Factory/Storage/StorageSyncSetFilterPacket.cs
+++ b/NebulaModel/Packets/Factory/Storage/StorageSyncSetFilterPacket.cs
@@ -1,11 +1,9 @@
-﻿namespace NebulaModel.Packets.Factory.Storage;
+namespace NebulaModel.Packets.Factory.Storage;
 
 public class StorageSyncSetFilterPacket
 {
     public StorageSyncSetFilterPacket() { }
 
-    // @TODO: This can potentially be moved up the call stack - in UI input - for storage boxes-
-    // to avoid sending redundant data like `StorageIndex` & `StorageType` per every grid slot.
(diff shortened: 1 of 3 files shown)


NebulaModel/Packets/Factory/Storage/StorageSyncSetFilterPacket.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace NebulaModel.Packets.Factory.Storage;
+
+public class StorageSyncSetFilterPacket
+{
+    public StorageSyncSetFilterPacket() { }
+
+    // @TODO: This can potentially be moved up the call stack - in UI input - for storage boxes-
+    // to avoid sending redundant data like `StorageIndex` & `StorageType` per every grid slot.
```
starfi5h, line 8: Yeah, but
1. The event is by the user and is seldom triggered, so there is no need to worry about bandwidth usage.
2. There are possibilities that other mods may directly call SetFilter without going the default UI route.
3. In the future if there are other buildings that use Storage (e.g. Battlefield Analysis Base), this code can be reused without adding patch to the UI.

Reousa: Understood! Removed the comment. 👍

starfi5h (approved):

Issue #470, opened 2021-11-06
NullReferenceException when player joins in PowerTowerUserLoadingRequestProcessor.ProcessPacket


hashworks:
![error](https://fb.hash.works/eGESft/)

The error occurred instantly when a player joined. We restarted multiple times, but the error kept occurring. The player then moved to the host planet while the error was active and the game was saved and restarted, "fixing" the error – but that might be unrelated.

It might happen [here](https://github.com/hubastard/nebula/blob/8d1a4e9b2350eda6a4b1af223618b6eddb499867/NebulaNetwork/PacketProcessors/Factory/PowerTower/PowerTowerUserLoadRequestProcessor.cs#L12) when either
* `packet` is null
* `factory.powerSystem` is null
* `factory.powerSystem.netPool` is null
* `GameMain.galaxy.PlanetById(packet.PlanetId).star` is null

hashworks:
Additional note: The joining player was in another system where the host wasn't before.

hashworks:
It seems like the player returning to the host planet "fixing" the error was unrelated.

PR #627, opened 2023-12-28
Rework power tower syncing

### LINKED ISSUE #448: Wireless charging effect persists on client after leaving the charge area
Just started playing recently (using latest version 0.6.0), my friend hosting and me joining the game. We've noticed today that the wireless charging effect (i.e. a large positive energy intake) from the wireless tower seems to persist on me (client) even after leaving the tower's area of effect - and it was up for a longer time (~5 minutes?).

Repro: join as a client, walk to a Wireless Power Tower to initiate mech charging, leave the tower area of effect, higlight mech core power indicator to see the charging value is still applied. Doesn't seem random at all - happened each and every time (at least since I've noticed this).

Let me know if I can provide any more details/logs/test this further.

starfi5h:
Replace the vanilla logic of power chargers and simply sync ON/OFF state.

PowerTowerManager now only maintains two pool, one is `LocalChargerIds` which stores all nodeIds of power chargers charging the local player. When add/remove nodeId from `LocalChargerIds`, it is broadcast to other players and stored in `RemoteChargerHashIds` as 64 bits hashIds, which planetId is the upper bits and nodeId is the lower bits.

If the player is in range of the charger, or `RemoteChargerHashIds` contains the id, then the charger will be set to working state, otherwise, it is set to idle state.

The pools are reset and recalculated whenever a player is disconnected to prevent persistent energy charges.
Fix #448 #470


PhantomGamers (approved):
Very nice!

NebulaPatcher/Patches/Transpilers/PowerSystem_Transpiler.cs
```diff
@@ -15,142 +17,196 @@ namespace NebulaPatcher.Patches.Transpilers;
...
+            return false;
+
+        if (GameMain.mainPlayer.planetId != powerSystem.factory.planetId)
+            return true;
+
+        return !Multiplayer.Session.PowerTowers.LocalChargerIds.Contains(nodeId);
+    }
+
+#pragma warning disable CA1868
+    private static void SetChargerRequriePower(PowerSystem powerSystem, int nodeId, float coreEnergyRatio)
```
PhantomGamers, line 150:
```suggestion
    private static void SetChargerRequirePower(PowerSystem powerSystem, int nodeId, float coreEnergyRatio)
```

Or should it be Requires/Required?

starfi5h: Given that it is updating `PowerNodeComponent.requiredEnergy`, Required may be better.

PR #634, opened 2024-01-04
Refactor networking logic & Deprecate `PlayerManager`

Reousa:
This is in preparation for implementing hot-swappable messaging libraries.

#### Things done:
- Move connection & disconnection logic out of `PlayerManager` to `Server`
- Move all messaging logic out of `PlayerManager` to `Server` & `Client` 
- Move player tracking out of `PlayerManager` to `Server` & add querying logic to an extension class
- Refactor `WebSocketService` outside of `Server`, and make it transparent to the rest of the program
- Deprecate & refactor all remaining methods inside `PlayerManager` to point to their counterparts

----

This PR is included in a series:
1. #634 - Working
2. #638 - Working
~~3. #639 - Working~~


NebulaAPI/Extensions/PlayerCollectionExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        return players
+            .Where(value => value.Connection.ConnectionStatus == status)
+            .ToList();
+    }
+
+    public static INebulaPlayer GetByPlayerId(
+        this IReadOnlyCollection<INebulaPlayer> players, ushort id)
+    {
+        return players.First(p => p.Id == id);
+    }
```
PhantomGamers, line 41: This will throw an exception if there is no player found with the given ID. Should we do this or do FirstOrDefault and just return null?

NebulaAPI/Extensions/PlayerCollectionExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+        this IReadOnlyCollection<INebulaPlayer> players, INebulaConnection conn)
+    {
+        return players.FirstOrDefault(p => p.Connection.Equals(conn));
+    }
+
+    public static INebulaPlayer GetByUsername(
+        this IReadOnlyCollection<INebulaPlayer> players, string username)
+    {
+        return players.First(p => p.Data.Username == username);
+    }
```
PhantomGamers, line 62: Here GetByConnectionHandle can return null but GetByUsername has the same behavior as GetByPlayerId where it will throw if the player is not found. So same question as before, is it better to throw here or return null? And shouldn't the behavior be the same for all of these?

Reousa: Fixed, all of them will now no longer throw

NebulaAPI/Extensions/IEnumerableExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace NebulaAPI.Extensions;
+
+public static class IEnumerableExtensions
+{
+    public static void ForEach<T>(this IEnumerable<T> source, Action<T> action)
```
Therzok, line 15: I'm not a big fan of this extension.

There are many allocations happening implicitly, and the JIT cannot optimize any of the work done because Action is also an allocated closure because captures are generally used.

The `IEnumerable` is always going to allocate an enumerator box, which is usually avoided for the usual collections (array, list, etc) if foreach is done directly, and in many places in the APIs in the language libraries.

I'm not 100% sure the APIs bring enough convenience - especially if accidentally used in hot loops - is worth it.

NebulaAPI/Extensions/PlayerCollectionExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        return players
+            .Where(value => value.Connection.ConnectionStatus == status)
+            .ToList();
+    }
+
+    public static INebulaPlayer GetByPlayerId(
+        this IReadOnlyCollection<INebulaPlayer> players, ushort id)
+    {
+        return players.FirstOrDefault(p => p.Id == id);
```
Therzok, line 29: this allocates a closure every time it's called.

Therzok: Any lambda in here that can't be marked static will likely allocate a closure, just a mention. That means that if it's called in hot code, it might trigger GCs.

Reousa:
@Therzok;
Extensions served as a temporary placeholder until I removed PlayerManager and conclusions were made over some discord discussions. 
Admittedly I hadn't considered performance impact yet, given the WIP nature of things, that would've been a later step. 😅

Adjustments made in any case, will rebase #638 after PR acceptance.

By extension & to keep reference, this PR also eliminates dependency on the `Locker` object, avoiding potential GC triggers in hotpaths from frequent `PlayerManager` member access. 

----
I sadly have no easy way to measure or compare these impacts as of yet, but a mocking of `Server` and `Client` to simulate some measurable traffic is something I'm attempting.

With that said, should pooling packets & potentially packet buffers be brought into the scope of this refactor (in a later PR)? 
I recon that'd have far more impact on easing out the GCs? What do you think?

Therzok:
Thanks for removing the allocs! Pooling sounds good, an alternative would be restructuring the generic packet handling to be use `where T:struct, IPacket` and the JIT will do everything else.

starfi5h (approved):

PR #638, opened 2024-01-06
Remove `PlayerManager` completely

Reousa:
Completely removes the now deprecated `PlayerManager` and updates all references to it's alternatives in `IServer`

----

This PR is included in a series:
1. #634 - Working
2. #638 - Working
~~3. #639 - Working~~

Closes #634


NebulaNetwork/PacketProcessors/Routers/PlanetBroadcastProcessor.cs
```diff
@@ -15,29 +15,20 @@ namespace NebulaNetwork.PacketProcessors.Routers;
...
     {
-        if (IsClient)
-        {
-            return;
-        }
+        //Forward packet to other users if we're the host
+        if (IsHost)
+            Multiplayer.Session.Server.SendToMatching(packet, p =>
+                p.Data.LocalPlanetId == packet.PlanetId &&
+                p.Connection.Equals(conn)
```
starfi5h, line 28: The condition is Not Equals to not send to the original sender

NebulaNetwork/PacketProcessors/Routers/StarBroadcastProcessor.cs
```diff
@@ -13,30 +13,22 @@
...
     {
-        if (IsClient)
-        {
-            return;
-        }
+        //Forward packet to other users if we're the host
+        if (IsHost)
+            Multiplayer.Session.Server.SendToMatching(packet, p =>
+                p.Data.LocalStarId == packet.StarId &&
+                p.Connection.Equals(conn)
```
starfi5h, line 28: The condition is Not Equals to not send to the original sender

NebulaAPI/Extensions/PlayerCollectionExtensions.cs
```diff
@@ -0,0 +1,147 @@
...
+
+    /// <summary>
+    /// Filters to connected players.
+    /// </summary>
+    /// <param name="players"></param>
+    /// <returns></returns>
+    public static IReadOnlyCollection<INebulaPlayer> Connected(
+        this IEnumerable<INebulaPlayer> players)
+    {
+        return players.FilterByConnectionStatus(EConnectionStatus.Connected);
```
starfi5h, line 125: Since Connected is a frequently used collection, consider caching and updating it only if there are EConnectionStatus changes? Otherwise `.ToList()` will be called multiple times to allocate memory.

Reousa: I plan to bring up some changes to be discussed that may ommit any reason to filter by connection status to begin with - by handling pending and syncing during the connection phase before adding the players to the list.

The performance impact of this is also very negligible considering the maximum amount of players per server can't be very high.

Until that's done or decided to not be done, think we could leave this part as is? Just to keep things simple until then.

starfi5h: It's not about the size, it's constantly and rapidly allocation and freeing memory will cause stutter from time to time in game when GC happen. CommonAPI has this kind of issue before.

Reousa: Aha gotcha, I get what you mean now. I went ahead and removed all allocative calls - it should in theory strictly return references now without allocating anything new.

If this is still unsatisfactory I'll add a cache for connection types as well.

starfi5h (approved):

PR #640, opened 2024-01-10
Implement gift command

mmjr-x:
Notable changes:
- Chat commands arguments now support double quoted grouping and first lvl escaping using \"
- Added gift command
- Made changes to the INetworkProvider interface to include a SendPacketToClient method and implemented it so it can send a packet from client to client without having to explicitly route it through the host (this is abstracted away)

TODO:
- Implement /gift <item> functionality
- Implement /gift energy functionality
- Implement gift success/failure mechanisms (to be able to refund the gift for various reasons)
- Support scientific (e.g. 1e3) and unit notations (1k) for quantity specification
- Add support for sending to a gift using id instead of username
- (Optional) Look how to get a list of connected users as a client (in order to validate the usernames before hand for the SendPacketToClient )

### DIFF
--- a/NebulaAPI/GameState/INetworkProvider.cs
+++ b/NebulaAPI/GameState/INetworkProvider.cs
@@ -46,5 +46,10 @@ public interface INetworkProvider : IDisposable
     /// </summary>
     void SendPacketToStarExclude<T>(T packet, int starId, INebulaConnection exclude) where T : class, new();
 
+    /// <summary>
+    ///     Send packet to Client directly (If possible) or indirectly by relaying via host (If Client is not directly reachable)
+    /// </summary>
+    void SendPacketToClient<T>(T packet, string clientUsername) where T : class, new ();
+
     void Update();
 }
--- a/NebulaModel/DataStructures/Chat/ChatCommandGiftType.cs
+++ b/NebulaModel/DataStructures/Chat/ChatCommandGiftType.cs
@@ -1,0 +1,8 @@
+﻿namespace NebulaModel.DataStructures.Chat;
+
+public enum ChatCommandGiftType
+{
+    Soil = 0, // TODO: Since Soil Pile (item id 1099 ) is also an item, we can probably remove this all together and just use Item
+    Item = 1,
+    Energy = 2
+}
--- a/NebulaModel/NetworkProvider.cs
+++ b/NebulaModel/NetworkProvider.cs
@@ -38,6 +38,9 @@ public abstract class NetworkProvider : INetworkProvider
     public abstract void SendPacketToStarExclude<T>(T packet, int starId, INebulaConnection exclude)
         where T : class, new();
 
+    public abstract void SendPacketToClient<T>(T packet, string clientUsername)
+        where T : class, new();
+
     public abstract void Update();
 
     public abstract void Start();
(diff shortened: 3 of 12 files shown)


PhantomGamers:
I know this is still marked as a draft but just to give my 2 cents with that latest commit, I don't love the idea of the command *requiring* a user id instead of a username.

I haven't done any work with these chat commands in the past but if usernames are not guaranteed to be unique, what happens with the whisper command if there are multiple people with the same username? Because that command does just take in a username and I think that's more intuitive for players to use, though I do fully support userid being an option.

I think maybe the best approach is to have username be the main method of sending things, and then if there are multiple people with the same name then when you execute the command the game would tell you there are multiple players with the same name and list out the ids of those players and ask the sender to specify which they were referring to.

Also smaller thing that I'm not sure if anyone else agrees with but IMO this command shouldn't be `/gift` but like `/give` or `/giveitem` or something like that.

mmjr-x:
@PhantomGamers How I implemented it now is that the command both works with userId or username (you can pick) if that is not the case at the moment then I made a mistake as it should definitely support both (and I will fix that), does that satisfy your request or is it more about the internals?

I am not sure what happens with the whisper command although I expect it just picks the first player matching that username in the list. Now that I think about It my code does the same if you provide a username that is shared by multiple users. I will make it so that it will block the gift in that case and will notify you to use the userId instead with the userIds for that username which are available to your, I think this falls in line with what you suggest?

I'm fine with anything regarding what the command should actually be called, I my mind I associated /give with just giving yourself something (as in a cheat command) so I though /gift would be more appropriate. However I can change it to /give if you want? (I could also change it to /give and make /gift an alias or vice versa (just like /g is)?

PhantomGamers:
> How I implemented it now is that the command both works with userId or username (you can use either) if that is not the case at the moment then I made a mistake as it should definitely support both (and I will fix that), does that satisfy your request or is it more about the internals?

I might have just misread the code then, I didn't actually try testing it in-game to see how it functioned, so yeah ignore if that was not relevant my bad.

> I am not sure what happens with the whisper command although I expect it just picks the first player matching that username in the list. Now that I think about It my code does the same if you provide a username that is shared by multiple users. I will make it so that it will block the gift in that case and will notify you to use the userId instead with the userIds for that username which are available to your, I think this falls in line with what you suggest?

Yeah I think this would be good, but I mean that doesn't necessarily have to happen in this commit. I think maybe these commands could be restructured so that that the interchangeability between the username and userid would just be available to any command that requires it. 

> I'm fine with anything regarding what the command should actually be called, I my mind I associated /give with just giving yourself something (as in a cheat command) so I though /gift would be more appropriate. However I can change it to /give if you want? (I could also change it to /give and make /gift an alias or vice versa (just like /g is))?

Yeah the alias idea could work. I understand what you mean about the cheat command though, I was thinking that too, but to me `/gift` isn't super intuitive as a command for giving items to other players, but yeah having both work could be good perhaps.

One more thing I just thought of but what happens if someone's username is just a number that could be a userid? Seems like this could be a potential problem as we;ll, though an edge case yeah.

R4wizard:
FWIF when I read the title of this PR, `/gift` immediately made me think of something like "claim your daily login reward" rather than an admin command to give things.

PR #650, opened 2024-01-30
Bugfix and enhancement

starfi5h:
image

- Fix error on `NebulaNetwork.Server.OnSocketDisconnection (NebulaAPI.Networking.INebulaConnection conn)`
- Fix error on `NebulaNetwork.SaveManager.LoadServerData ()`
- Fix error on StationUIManager.UpdateStorage when station.storage is null.
- When client views a star system in starmap mode, let it query the server to ask for potential dyson sphere data.
- Add config `General`:`NameTagSize` to scale other player nametag size. Default value is 100
- Add config `General`:`Enable Achievement` to enable achievement in multiplayer game.

### DIFF
--- a/NebulaNetwork/Messaging/WebSocketService.cs
+++ b/NebulaNetwork/Messaging/WebSocketService.cs
@@ -57,8 +57,12 @@ public class WebSocketService : WebSocketBehavior
 
     protected override void OnClose(CloseEventArgs e)
     {
-        var connection = connections[Context.UserEndPoint.GetHashCode()];
-        connections.Remove(connection.GetHashCode());
+        if (!connections.TryGetValue(Context.UserEndPoint.GetHashCode(), out var connection))
+        {
+            return;
+        }
+        connections.Remove(Context.UserEndPoint.GetHashCode());
+
         // If the reason of a client disconnect is because we are still loading the game,
         // we don't need to inform the other clients since the disconnected client never
         // joined the game in the first place.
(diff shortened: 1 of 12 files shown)


NebulaNetwork/Messaging/WebSocketService.cs
```diff
@@ -81,19 +85,22 @@ protected override void OnClose(CloseEventArgs e)
 
     protected override void OnError(ErrorEventArgs e)
     {
+        if (!connections.TryGetValue(Context.UserEndPoint.GetHashCode(), out var connection))
+        {
+            return;
+        }
```
PhantomGamers, line 91: Why is this needed?

According to https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.remove?view=netstandard-2.0, remove() returns true if the element is found and removed or false if it is not found, so in this case the TryGetValue should just be redundant and result in an additional lookup.

NebulaNetwork/PacketProcessors/Universe/DysonSphereRequestProcessor.cs
```diff
@@ -60,6 +60,29 @@ protected override void ProcessPacket(DysonSphereLoadRequest packet, NebulaConne
             case DysonSphereRequestEvent.Unload:
                 Multiplayer.Session.DysonSpheres.UnRegisterPlayer(conn, packet.StarIndex);
                 break;
+
+            case DysonSphereRequestEvent.Query:
+                // Ingroe query if dyson sphere doesn't exist on host
```
PhantomGamers, line 65:
```suggestion
                // Ignore query if dyson sphere doesn't exist on host
```

PR #653, opened 2024-02-17
Rework drone syncing for 0.10.29.21950

starfi5h:
Using a simpler implementation for drone syncing.

- Sync `Prebuild.itemRequired` change event that the item is taken from the player package in `ConstructionModuleComponent.PlaceItems`. When the server receives and the prebuild item is satisfied, refund the item.

- Manage drone and craft data from other players in a separate pool from the vanilla pool to avoid interfering with the vanilla save data. The remote drones will use the mimic function of `ConstructionSystem.UpdateDrones` to update position and state. Its only real interaction is repairing the damaged Entity. The rest is just for updating the animation model.

- Since the drones now go to the nearest target first, the build plan coordination can be ignored without losing too much efficiency. `BuildFinally` has already been guarded to prevent duplicate build calls.


NebulaWorld/Player/DroneManager.cs
```diff
@@ -2,187 +2,72 @@
...
+    private readonly List<CraftData> crafts = [];
+    private readonly Stack<int> craftRecyleIds = [];
+    private readonly DataPool<DroneComponent> drones = new();
 
     public DroneManager()
     {
-        PlayerDroneBuildingPlans = [];
-        PendingBuildRequests = [];
-        CachedPositions = [];
         lastCheckedTick = 0;
```
Zzarek, line 26: redundant?

NebulaWorld/Player/DroneManager.cs
```diff
@@ -192,181 +77,215 @@ public static void RefreshCachedPositions()
...
+            if (result != 0 && (ptr.stage == 2 || ptr.stage == 3 || ptr.stage == 4))
+            {
+                ptr.movement--;
+                if (ptr.movement <= 0 || flag2)
+                {
+                    ptr.movement = 0;
+                    ptr.stage = 4;
+                    ptr.targetObjectId = 0;
+                }
+                else if (ptr.nextTarget1ObjectId != 0)
```
Zzarek, line 172: This really has a refactor me feel, but I dont have the full context. Seems like the entire conditional set belongs in a method. 

Also, the 1, 2 ,3 seems like you are either following DSP values, or 1,2, 3 are specifically coded. Hard to tell withotu a deep dive.

starfi5h: As the comment state, this function is a modified version of `ConstructionSystem.UpdateDrones()`.
Those 1,2,3 are DSP values used in `DroneComponent.InternalUpdate()`.

Zzarek (approved):

sp00ktober (approved):

PR #660, opened 2024-03-14
Fix various bugs

starfi5h:
Fix:
- Fix `UIPlanetDetail` error when Client joins from lobby
- Fix `PropertyLogic.UpdateProduction` error in Client
- Fix error when reversing belts with items on them
- Fix error when removing base pit on foundations
- Fix index out-of-range error in `DFGFormationAddUnitProcessor`
- Temporary fix NRE of factoryStatPool in PowerSystem.GameTick (root cause in unknown)

Code Changes:
- Recycle all fleets of Client when joining to prevent `CombatModuleComponent` error
- Show warning when attempting to send unregistered packet type
- (Client) Preset `enemyId` now will kill the enemy first if it occupies the incoming id


NebulaModel/Packets/Factory/Belt/BeltReverseRequest.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NebulaModel.Packets.Factory.Belt;
+
+public class BeltReverseRequest
+{
+    public BeltReverseRequest() { }
+
+    public BeltReverseRequest(int beltId, int planetId, int authorId)
```
PhantomGamers, line 7: This should be `BeltReverseRequestPacket` (the filename as well) to fit in with the rest of the files in that section

NebulaNetwork/PacketProcessors/Factory/Belt/BeltReverseProcessor.cs
```diff
@@ -1,47 +1,62 @@
...
+        }
+
         using (NebulaModAPI.MultiplayerSession.Factories.IsIncomingRequest.On())
         {
             NebulaModAPI.MultiplayerSession.Factories.EventFactory = factory;
             NebulaModAPI.MultiplayerSession.Factories.TargetPlanet = packet.PlanetId;
-            if (NebulaModAPI.MultiplayerSession.LocalPlayer.IsHost)
-            {
-                // Load planet model
-                NebulaModAPI.MultiplayerSession.Factories.AddPlanetTimer(packet.PlanetId);
```
PhantomGamers, line 30: Is this not needed anymore?

starfi5h: Currently the function that use `planet.physics` is `CargoTraffic.AlterBeltConnections()`, which the part is guarded by `this.isLocalLoaded (= this.planet.factoryLoaded)`. Therefore I think it's safe to remove the PlanetTimer, but I can add back if it is more robust.

PhantomGamers: Hm, I think originally it wasn't guarded because I believe I added this due to errors that were occurring, but with this guard then the belt connections will not be modified if the planet is not loaded right? So we would still want to load it?

starfi5h: The guarded part is about modifying rendering and collider variables, if my understanding is correct.
But I'll add back just in case.

PhantomGamers (approved):

PR #664, opened 2024-03-21
Korean Display in Chat

sk7725:
Notable Changes:

- Added Korean font and TextMeshPro fallback
  - Font is Pretendard (which has no licensing issues and can be used commercially)
  - TMP atlas is "Pretendard-Regular"
- The default behaviour is a dynamic atlas, but I've also included a static atlas (lower quality) just in case of issues.
  - The static atlas (Pretendard-Static) is currently unused.


sk7725:
Will convert to a non-draft PR tomorrow with multi-user testing tomorrow

starfi5h (approved):

PR #666, opened 2024-03-23
Construction drones range limit & command /playerdata

starfi5h:
## Construction drones range limit
Make construction drones only launch if the local player is closer than other remote players, or within 15m.
So that the drones won't try to fly to far prebuild that will be built by other players first.

## SaveManager
Clear playerSaves from the previous session when server starts.  

Add new chat command `/playerdata`.
playerdata
- `/playerdata list`: list all stored player data in format [hash] playerName
- `/playerdata remove <hash>`: remove the target player data from SaveManager.
- `/playerdata load <hash>`: load the target player data to overwrite inventory and appearance of local player, then teleport to the stored location.

With this command, it can achieve the following goal for the host:
1. Reset player data so they can join as a new player if they choose the wrong planet to start in the lobby.
2. After transferring the save files, get the stored player data instead of the data from the previous host.

### DIFF
--- a/NebulaNetwork/SaveManager.cs
+++ b/NebulaNetwork/SaveManager.cs
@@ -91,25 +91,31 @@ public static class SaveManager
         File.Move(str1, str2);
     }
 
-    public static void LoadServerData()
+    public static void LoadServerData(bool loadSaveFile)
     {
-        var path = GameConfig.gameSaveFolder + DSPGame.LoadFile + FILE_EXTENSION;
+        playerSaves.Clear();
 
-        // var playerManager = Multiplayer.Session.Network.PlayerManager;
-        if (!File.Exists(path))
+        if (!loadSaveFile)
         {
             return;
         }
+        var path = GameConfig.gameSaveFolder + DSPGame.LoadFile + FILE_EXTENSION;
+        if (!File.Exists(path))
+        {
+            Log.Info($"No server file");
+            return;
+        }
 
-        var source = File.ReadAllBytes(path);
-        var netDataReader = new NetDataReader(source);
-        ushort revision;
         try
         {
+            var source = File.ReadAllBytes(path);
+            var netDataReader = new NetDataReader(source);
+            ushort revision;
+
             var revString = netDataReader.GetString();
             if (revString != "REV")
             {
-                throw new Exception();
+                throw new Exception("Incorrect header");
             }
 
             revision = netDataReader.GetUShort();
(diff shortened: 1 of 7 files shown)


NebulaNetwork/SaveManager.cs
```diff
@@ -119,43 +125,45 @@ public static void LoadServerData()
                 // Supported revision: 5~8
                 if (revision is < 5 or > REVISION)
                 {
-                    throw new Exception();
+                    throw new Exception($"Unsupport version {revision}");
```
PhantomGamers, line 128:
```suggestion
                    throw new Exception($"Unsupported version {revision}");
```

NebulaNetwork/SaveManager.cs
```diff
@@ -119,43 +125,45 @@ public static void LoadServerData()
...
+                else if (playerData == null)
+                {
+                    Log.Warn($"Could not load player data from unsupported save file revision {revision}");
+                }
             }
         }
+        catch (Exception e)
+        {
+            playerSaves.Clear();
+            Log.WarnInform("Skipping server data due to following excpetion:\n" + e.Message);
```
PhantomGamers, line 163:
```suggestion
            Log.WarnInform("Skipping server data due to exception:\n" + e.Message);
```
or "the following exception" but maybe that's too long

NebulaWorld/Chat/Commands/PlayerDataCommandHandler.cs
```diff
@@ -0,0 +1,167 @@
...
+public class PlayerDataCommandHandler : IChatCommandHandler
+{
+    public void Execute(ChatWindow window, string[] parameters)
+    {
+        if (parameters.Length < 1)
+        {
+            throw new ChatCommandUsageException("Not enough arguments!".Translate());
+        }
+
+        // Due to dependency order, get SaveManager.playerSaves by reflection
```
PhantomGamers, line 28: This is fine for now but I was actually just thinking that maybe it would make more sense if SaveManager was in NebulaWorld instead of NebulaNetwork. I guess that would solve this issue too without needing reflection, though I'm not sure if there'd be any issues with that.

PhantomGamers (changes requested):
Are these playerdata commands only available for the host/authenticated clients? If not, should they be limited?

Also not a blocking issue but:
> After transferring the save files, get the stored player data instead of the data from the previous host.
So once this is done, the player data from the previous host is permanently lost? If the save is  transferred back then the original host would be stuck with the other host's player data, and the other host's player data doesn't get updated to match what was done on the main data?

I think the best solution would be to just have the host's data also stored in the .server file with a hash (so the host would also generate a key file), and then swapping back and forth would be possible.

NebulaWorld/Chat/Commands/PlayerDataCommandHandler.cs
starfi5h: If SaveManager is in NebulaWorld, then `SaveManager.LoadServerData` will not able to be called in `Server.Start()`. It will have to hook in other place.
Hmm.. I guess this is fine and can further decouple the part in Server?

PhantomGamers: That would still work anyway since NebulaNetwork references NebulaWorld, no?

starfi5h:
> Are these playerdata commands only available for the host/authenticated clients? If not, should they be limited?

Currently it only has effects in host, as client will see 0 player data in SaveManager.
I'm not sure if it should be open to clients.

> Also not a blocking issue but:
> 
> > After transferring the save files, get the stored player data instead of the data from the previous host.
> 
> So once this is done, the player data from the previous host is permanently lost? If the save is transferred back then the original host would be stuck with the other host's player data, and the other host's player data doesn't get updated to match what was done on the main data?

No. It's not lost. As in [SaveManager.SaveServerData()](https://github.com/NebulaModTeam/nebula/blob/master/NebulaNetwork/SaveManager.cs#L25-L56), it will store all the hash-IPlayerData pair in `playerSaves` and the host hash-IPlayerData in the `.server` file.   
When a `/playerdata load` command is used, it just copys the inventory and appearance, the source pair is still in the SaveManager. So as long as the original host's `player.key` file remains the same. it can login to get back the data.

> I think the best solution would be to just have the host's data also stored in the .server file with a hash (so the host would also generate a key file), and then swapping back and forth would be possible.

It's already implemented in https://github.com/NebulaModTeam/nebula/blob/master/NebulaNetwork/SaveManager.cs#L45-L47

NebulaWorld/Chat/Commands/PlayerDataCommandHandler.cs
starfi5h: oh, my bad. I've mistaken the order

PhantomGamers (approved):

PR #682, opened 2024-06-19
Sync Dark Fog communicator and bugfix 

starfi5h:
New Features:
- Sync Dark Fog communicator (set aggressive or truce)
Dark Fog communicator is now available in multiplayer. It will not cost metadata as it is in sandbox mode.
- Show server last save time in client esc menu
LastSaveTime

Bugfix:
- Fix `NgrokManager.IsNgrokActive` crash when calling `Log.Error` on non-main thread
- Fix veins don't get buried by foundations on remote planets

### DIFF
--- a/NebulaNetwork/Ngrok/NgrokManager.cs
+++ b/NebulaNetwork/Ngrok/NgrokManager.cs
@@ -298,9 +298,8 @@ public class NgrokManager
             _ngrokProcess?.Refresh();
             return !_ngrokProcess?.HasExited ?? false;
         }
-        catch (Exception e)
+        catch
         {
-            Log.Error(e);
             return false;
         }
     }
--- a/NebulaNetwork/Ngrok/NgrokManager.cs
+++ b/NebulaNetwork/Ngrok/NgrokManager.cs
@@ -298,8 +298,9 @@ public class NgrokManager
             _ngrokProcess?.Refresh();
             return !_ngrokProcess?.HasExited ?? false;
         }
-        catch
+        catch (Exception e)
         {
+            Log.Error(e);
             return false;
         }
     }
--- a/NebulaNetwork/PacketProcessors/Factory/Foundation/VeinPosUpdateProcessor.cs
+++ b/NebulaNetwork/PacketProcessors/Factory/Foundation/VeinPosUpdateProcessor.cs
@@ -1,0 +1,61 @@
+﻿#region
+
+using NebulaAPI.DataStructures;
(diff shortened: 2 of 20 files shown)


NebulaNetwork/Ngrok/NgrokManager.cs
```diff
@@ -298,9 +298,8 @@ public bool IsNgrokActive()
             _ngrokProcess?.Refresh();
             return !_ngrokProcess?.HasExited ?? false;
         }
-        catch (Exception e)
+        catch
         {
-            Log.Error(e);
```
PhantomGamers, line 303: Do we not want to log cases where there is an error?

NebulaNetwork/PacketProcessors/Factory/Foundation/VeinPosUpdateProcessor.cs
```diff
@@ -0,0 +1,61 @@
...
+        if (planet != GameMain.localPlanet)
+        {
+            return;
+        }
+
+        // Update GPU models on the local planet
+        var rot = Maths.SphericalRotation(pos, Random.value * 360f);
+        GameMain.gpuiManager.AlterModel(veinData.modelIndex, veinData.modelId, packet.VeinId, pos, rot, false);
+        var veinProto = LDB.veins.Select((int)veinData.type);
+        if (veinProto != null)
```
PhantomGamers, line 38: Early return instead?
```
if (veinProto == null)
{
	return;
}
```

NebulaNetwork/Ngrok/NgrokManager.cs
starfi5h: Does logging of [InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=netframework-4.7.2) of `Process.HasExited` really help debugging?
Wouldn't [NgrokLastErrorCode](https://github.com/NebulaModTeam/nebula/blob/master/NebulaNetwork/Ngrok/NgrokManager.cs#L120-L124) more helpful?

PhantomGamers: Well the ngroklasterrorcode will only get populated if the error originates from the ngrok process directly. This check was because of ppl getting errors/crashes/etc because the process wasn't started properly at all (e.g. anti-virus quarantined the ngrok exe)

PhantomGamers (approved):
looks good, but id recommend bumping version and changelog before merging

PR #688, opened 2024-07-03
Removes the requirement of Steam on Headless Server, also a Object Reference fix

AlienXAXS:
This PR adds the needed harmony patches to no longer require Steam for loading the headless server.  This gets rid of the need for server owners to either run steam, or install some Steam Emulator such as Goldberg's, which is what most people are doing now...

It also fixes a Object Reference error during Headless load on UICommunicatorIndicator as there is no UI on load of a headless.

This is my first PR for this project, I hope that everything is in the right places and explained enough - apologies if not.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/Dedicated_Server_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/Dedicated_Server_Patch.cs
@@ -3,11 +3,13 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using System.Reflection.Emit;
 using HarmonyLib;
 using NebulaModel;
 using NebulaModel.Logger;
 using NebulaWorld;
+using Steamworks;
 using UnityEngine;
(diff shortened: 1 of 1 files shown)


PhantomGamers:
I don't really love this. Ideally yes people would be able to use the dedicated server functionality without Steam installed or running, or requiring an emulator like Goldberg, but this would be bypassing drm directly in nebula which is both legally and morally questionable.

If we could somehow get an OK from Youthcat to implement this I wouldn't have a problem but I'm not sure how likely that is to happen.

AlienXAXS:
Hi Phantom,
I thought that this may well be the case, and I understand it's... shall we say grey area.   Thankfully the code base only applies these patches when -server is in the argument list, perhaps that is something that is useful for them to approve such a PR.

Either way, the UI fix should be done regardless of the NoSteam part :)

PhantomGamers:
So I'll close this for now, but if you want to resubmit a pr with only the UI fix that would definitely get merged :)

PR #690, opened 2024-07-04
No longer initialise DiscordManager on a headless server.

AlienXAXS:
This stops a skippable error during headless startup too, there is no need to init discord here as the likelihood of it being present is extremely slim.

I've gone through DiscordManager and it looks as though no extra client == null checks are required.


PhantomGamers:
There are certainly cases where you'd want to use the discord integration with headless mode. If you don't, disabling it in the settings is an option

PhantomGamers:
Also just an alternative solution if just disabling it in the config isn't optimal, perhaps another argument could be added to disable discord integration that way. `-no-discord` or something

PR #692, opened 2024-07-17
Headless now calculates planetary shields on CPU

AlienXAXS:
Currently planet shields do nothing on Headless as they are emptied out and pretty much turned off as it's not possible to simulate how much energy is inside of them as `energy` is set to zero as well as `isEmpty` being set to True.

You can start reading more about this here: https://github.com/NebulaModTeam/nebula/issues/668#issuecomment-2227037945

While I totally understand putting 0.95 here for the headless server to assume that the shield is always attempting to be charged to 95% capacity can be a little bit of a horrible fix, I believe that this is function over form.  I'm sure many of the players running headless would prefer their shields to work as intended, instead of doing nothing.

Running headless also causes a large desync issue with landing relays.  For example if a relay attempts to land on a shielded planet your client will quite rightly dismiss the incoming relay and send it home, but on headless the relay still lands and creates a base.
A handful of times now I have had invisible enemies blowing up my shielded planets structures, just to do a `/r` to see that a enemy base actually exists and is chewing through my buildings.

I noticed during my testing that the RecalculatePhysicsShape method is no longer called once the variable (0.95 in this instance) became a full 1.0f, this value was read-in from the ComputeShader for PlanetAT's.   Setting it to 0.95 seemed to allow the method to be called enough so that the energyMaxTarget was updated quickly enough during the game.

I have been running this patch on my own save for the last 4 days, the instance has also been running while no one is online and not a single relay has managed to land on my planets that have active shields.

### DIFF
--- a/NebulaPatcher/Patches/Dynamic/Dedicated_Server_Patch.cs
+++ b/NebulaPatcher/Patches/Dynamic/Dedicated_Server_Patch.cs
@@ -171,15 +171,47 @@ internal class Dedicated_Server_Patch
         return false;
     }
 
-    [HarmonyPostfix]
+    [HarmonyPrefix]
     [HarmonyPatch(typeof(PlanetATField), nameof(PlanetATField.RecalculatePhysicsShape))]
-    public static void RecalculatePhysicsShape_Postfix(PlanetATField __instance)
+    public static bool RecalculatePhysicsShape_Prefix(PlanetATField __instance)
     {
-        // vanilla use GPU to calculate the shape of shield that is not fully covered the whole planet
-        // In this patch it will act as it has been fully covered to make the planet shield effective
-        if (!__instance.isEmpty)
+        // If we're the server, let's update the planet shields manually.
+        // This is required as checks to planetary shields fail, as there is no GPU to simulate the shields
+        // they become `isEmpty = true`.  This causes some checks to fail such as:
+        //  - Relays still landing on a planet when a shield is online
+        //  - Some space to surface weaponry ignoring the shields
+        if (Multiplayer.IsDedicated)
         {
+            if (__instance.generatorCount == 0)
+            {
+                __instance.ClearPhysics();
+                __instance.energyMaxTarget = 0L;
+            }
+
+            __instance.CreatePhysics();
             __instance.isSpherical = true;
+
+            /*
+             * This is usually computed on the GPU, No idea what math the GPU does though, so we're shoving 0.95 here.
+             * On my own testing this goes as low as 0.35 during initial planet shield spin up
+             * and when it hits 1.0 it seems to stop calling this method.
+             */
+            __instance.energyMaxTarget = (long)(1200000000000.0 * 0.95 + 0.5);
+
+            if (__instance.energy > 0)
+            {
+                __instance.isEmpty = false;
+            }
+
+            // I believe these are used in raycasting tests to see if a relay would hit a shield, so we need them.
+            if (__instance.colliderHotTicks > 0)
+                __instance.OpenColliderObject();
+            else
+                __instance.CloseColliderObject();
+
+            return false;
         }
+
+        return true;
     }
 }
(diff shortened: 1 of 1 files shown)


NebulaPatcher/Patches/Dynamic/Dedicated_Server_Patch.cs
```diff
@@ -171,15 +171,47 @@ public static bool UICommunicatorIndicatorOnLateUpdate_Prefix()
...
     {
-        // vanilla use GPU to calculate the shape of shield that is not fully covered the whole planet
-        // In this patch it will act as it has been fully covered to make the planet shield effective
-        if (!__instance.isEmpty)
+        // If we're the server, let's update the planet shields manually.
+        // This is required as checks to planetary shields fail, as there is no GPU to simulate the shields
+        // they become `isEmpty = true`.  This causes some checks to fail such as:
+        //  - Relays still landing on a planet when a shield is online
+        //  - Some space to surface weaponry ignoring the shields
+        if (Multiplayer.IsDedicated)
```
starfi5h, line 183: There is no need for testing Multiplayer.IsDedicated as Dedicated_Server_Patch only patches in dedicated server mode.

AlienXAXS: Oh yes, my apologies - it has been removed.

AlienXAXS:
Just as a side note for further discussions:
I wonder if it may be worth syncing when the server decides to send a relay back to the hive (Direction on the DFRelayComponent gets set to -1 for this), so that this is a properly synced event?  I ask because relays will still attempt to land on a planet if you are not in the same galaxy as the relay.

I was tempted to have a go at doing that, but I wanted to ask if it's something you guys feel is needed first.

starfi5h:
Are you saying that `ArriveBase()` make relay behavior differently because `TestRelayCondition` or `CheckLandCondition` gives different results? Hmm.. that should be dealt with.
I would imagine a proper way is to let server check all the conditions and client only updates the attributes of the relay or create/connect to the base. I can work on it if you have a more interesting topic to do.

Here is guide for transpiler: https://github.com/BepInEx/HarmonyX/wiki/Transpiler-helpers

AlienXAXS:
Well for example, running two instances (both graphical) on a world where a relay is approaching a planet with shields on the current Nebula build will in my testing cause the host to ping the relay back to the hive dock, but the client will have the relay land and build a base on the planet.

Adding more sync to whenever the Direction variable is changed would fix a lot of these problems.

ArriveBase on headless:
It's possible that ArriveBase will take a different codepath running on Headless than when running on a system that is able to run ComputeShaders due to the `TestRelayCondition` test in theory being reversed compared to the connected client.

Areas to Sync from within DFRelayComponent.RelaySailLogic:
Line 898's if statement:
```csharp
if ((vectorLF4 - planetFactory.planet.uPosition).sqrMagnitude < 360000.0)
{
    this.targetLPos = Vector3.zero;
    this.targetYaw = 0f;
    this.baseState = 0;
    this.baseId = 0;
    this.baseTicks = 0;
    this.baseEvolve = default(EvolveData);
    this.baseRespawnCD = 0;
    this.direction = -1;
    this.param0 = 0f;
}
```

Same for the if statement starting line 919.
(This is where PlanetAT is checked and the relay is sent home if a planet is shielded)
```csharp
else if (planetFactory != null && planetFactory.planetATField != null && !planetFactory.planetATField.TestRelayCondition(this.targetLPos))
{
	flag = true;
	this.targetAstroId = 0;
	this.targetLPos = Vector3.zero;
	this.targetYaw = 0f;
	this.baseState = 0;
	this.baseId = 0;
	this.baseTicks = 0;
	this.baseEvolve = default(EvolveData);
	this.baseRespawnCD = 0;
	this.direction = -1;
	this.param0 = 0f;
}
```

Also:
- DFRelayComponent.LeaveBase being called
- DFRelayComponent.ArriveBase's flag being NOT set will also return the relay back to the hive. (notice line 562 -> 567)

I am more than happy for you to take this PR and throw it into your own merge, I think that it would take me a long time to figure out transpliers, even with that article which ive already read before.

My next task is the online playerlist on holding Tab, and also trying to figure out why Energy Exchangers when in Charge mode dont work for connected clients, but the host is still making the charged batteries.

starfi5h (approved):

PR #697, opened 2024-07-22
Implemented a graphical player list

AlienXAXS:
Hotkey is configurable via the usual in-game settings screen.  Closes https://github.com/NebulaModTeam/nebula/issues/390
UI Screen will display when the user holds Tab (configured by default to be Tab), much like any other FPS style game.

It may be possible to add more information to it later, when some more syncing is complete.

image

image

### DIFF
--- a/NebulaModel/MultiplayerOptions.cs
+++ b/NebulaModel/MultiplayerOptions.cs
@@ -131,7 +131,7 @@ public class MultiplayerOptions : ICloneable
 
     [DisplayName("Player List Hotkey")]
     [Description("Keyboard shortcut to display the Connected Players Window")]
-    public KeyboardShortcut PlayerListHotkey { get; set; } = new(KeyCode.Tab);
+    public KeyboardShortcut PlayerListHotkey { get; set; } = new(KeyCode.F1);
 
     [DisplayName("Auto Open Chat")]
     [Category("Chat")]
--- a/NebulaModel/MultiplayerOptions.cs
+++ b/NebulaModel/MultiplayerOptions.cs
@@ -131,7 +131,7 @@ public class MultiplayerOptions : ICloneable
 
     [DisplayName("Player List Hotkey")]
     [Description("Keyboard shortcut to display the Connected Players Window")]
-    public KeyboardShortcut PlayerListHotkey { get; set; } = new(KeyCode.F1);
+    public KeyboardShortcut PlayerListHotkey { get; set; } = new(KeyCode.BackQuote);
 
     [DisplayName("Auto Open Chat")]
     [Category("Chat")]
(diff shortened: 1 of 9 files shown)


NebulaWorld/UIPlayerList/UIPlayerWindow.cs
```diff
@@ -110,7 +110,7 @@ public void WindowHandler(int id)
                             var pName = player.Username;
                             var pDistanceFromLocalPlayer = "";
                             var pLocation = "";
-
+
```
AlienXAXS, line 113: Man Visual Studio keeps doing this and it's sure bugging me.

NebulaWorld/UIPlayerList/UIStyles/DialogStyles.cs
```diff
@@ -0,0 +1,42 @@
...
+        // ReSharper disable once InconsistentNaming
+        private static Texture2D _textureMemory = null;
+
+        public static GUIStyle WindowBackgroundStyle()
+        {
+            var width = 32;
+            var height = 32;
+
+            if (_textureMemory == null)
+                _textureMemory = MakeTex(width, height, new Color32(36, 67, 76, 200);
```
starfi5h, line 16: Error	CS1026	) expected

NebulaModel/MultiplayerOptions.cs
```diff
@@ -129,6 +129,10 @@ public bool StreamerMode
     [Description("Keyboard shortcut to toggle the chat window")]
     public KeyboardShortcut ChatHotkey { get; set; } = new(KeyCode.BackQuote, KeyCode.LeftAlt);
 
+    [DisplayName("Player List Hotkey")]
+    [Description("Keyboard shortcut to display the Connected Players Window")]
+    public KeyboardShortcut PlayerListHotkey { get; set; } = new(KeyCode.Tab);
```
starfi5h, line 134: No sure if I'm a fan of setting default key as `Tab`. The key is used in many places in default keybinds.
It would be quite annoying if the window kept popping out when doing other operations.

starfi5h (changes requested):

NebulaWorld/UIPlayerList/UIStyles/DialogStyles.cs
AlienXAXS: I sure am curious to how i even built this with that missing... but It's fixed now.

NebulaModel/MultiplayerOptions.cs
AlienXAXS: - Changed to F1.
- No longer shows if escScreen, Techtree or dysonEditor windows are open

starfi5h (approved):

PR #743, opened 2025-10-22
Refactor CommandLineOptions and bugfix

starfi5h:
### Changes
- Change dedicated server option from "-server" to "-dsp_server"
This is because the -server argument is already used in Unity 2022 version.
- Add new command line option "-newgame-default" to start game with vanilla default settings. 
- Change error message format. It can now show the extra info of the mod ErrorAnalyzer.

### Bugfix
- Fix bug that ray receiver power request values are not correct in multiplayer.

### Refactor CommandLineOptions

- Separate command line arguments handling from `NebulaPlugin`
- Add new class `CommandLineOptions` and `GameDescSettings` to deal with the related logic

### DIFF
--- a/NebulaModel/CommandLineOptions.cs
+++ b/NebulaModel/CommandLineOptions.cs
@@ -40,8 +40,12 @@ public class CommandLineOptions
                     ParseNewGameArguments(args, i);
                     break;
 
+                case "-newgame-default":
+                    ParseNewGameConfigArguments(useModConfigFile: false);
+                    break;
+
                 case "-newgame-cfg":
-                    ParseNewGameConfigArguments();
+                    ParseNewGameConfigArguments(useModConfigFile: true);
                     break;
 
                 case "-load":
@@ -116,7 +120,7 @@ public class CommandLineOptions
         return false;
     }
 
-    private bool ParseNewGameConfigArguments()
+    private bool ParseNewGameConfigArguments(bool useModConfigFile)
     {
         NewGameArgumentExists = true;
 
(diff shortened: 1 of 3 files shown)


NebulaModel/CommandLineOptions.cs
```diff
@@ -0,0 +1,221 @@
...
+        NewGameArgumentExists = true;
+        if (currentIndex + 3 < args.Length)
+        {
+            if (!int.TryParse(args[currentIndex + 1], out var seed))
+            {
+                Log.Warn($">> Can't set galaxy seed: {args[currentIndex + 1]} is not a integer");
+            }
+            else if (!int.TryParse(args[currentIndex + 2], out var starCount))
+            {
+                Log.Warn($">> Can't set star count: {args[currentIndex + 2]} is not a integer");
```
PhantomGamers:
```suggestion
                Log.Warn($">> Can't set star count: {args[currentIndex + 2]} is not an integer");
```

NebulaModel/GameDescSettings.cs
```diff
@@ -0,0 +1,66 @@
...
+
+        var resourceMultiplier = customFile.Bind("Basic", "resourceMultiplier", -1f,
+            "Resource Multiplier. Infinite = 100. Negative value: Default(1.0f) or remain the same.").Value;
+        if (resourceMultiplier >= 0f)
+        {
+            gameDesc.resourceMultiplier = resourceMultiplier;
+        }
+
+        gameDesc.isPeaceMode = customFile.Bind("General", "isPeaceMode", false,
+            "False: Enable enemy force (combat mode)").Value;
```
PhantomGamers:
```suggestion
            "False: Enable enemy forces (combat mode)").Value;
```

PhantomGamers (approved):

PR #776, opened 2026-09-09
Synchronize wireless power tower charging

Linwendong83:
Wireless power towers currently do not synchronize their charging state in multiplayer. When a remote player starts or stops charging, the tower's power demand and UI state can become inconsistent between the host and clients.

This change tracks charging membership per player and planet. Clients send snapshots of the towers they are currently using; the host validates them, keeps the authoritative state, and broadcasts updates to other players. Joining clients receive the current snapshot.

Power demand is enabled while at least one player is using a tower and is not multiplied by the player count. Energy is still applied only to the local mecha when that player is actually within charging range, preserving the original single-player behavior.

Charging state is cleared when a player changes planets or disconnects, when charging is no longer needed, and when a tower is removed. The power-node UI uses the synchronized player count.

The README checkbox is updated, and tests cover snapshot serialization, shared towers, cleanup paths, charger behavior, and transpiler failure handling.


Linwendong83:
Closing as these changes are maintained downstream in our fork.

